PR #20, opened 2024-06-13
added Dockerfile

xinredhat:
1. Added a Dockerfile for running CI jobs
2. Added yamllint.yaml
3. Downgrade golang to 1.21


ci/config/yamllint.yaml
otaviof, line 1: Could we move this file to the project root? So, when using `yamllint` outside of CI relies on the same configuration file?

xinredhat: sure, then we move the both files `yamlint.yaml` and `Dockerfile` to the project root, WDYT?

xinredhat: No, I won't move the `Dockerfile` to the project root, as it's only used for CI

xinredhat: fixed

otaviof (approved):

PR #42, opened 2024-06-22
RHTAPINST-45: ArgoCD Declarative Token Creation

otaviof:
Replacing the `backing-services`'s `post-deploy.sh` hook with a Kubernetes native approach, using the same container images employed on other resources and ArgoCD itself.

See also https://github.com/argoproj/argo-cd/issues/9884


charts/rhtap-backing-services/templates/NOTES.txt
```diff
@@ -1,13 +1,23 @@
...
 {{- end }}
+
+{{- if .Values.backingServices.argoCD.enabled }}
+  {{- $argoCD := .Values.backingServices.argoCD }}
+ArgoCD:
+  - Name: {{ $argoCD.name }}
+    Namespace: {{ $argoCD.namespace }}
+    Server URL: {{ include "backingServices.argoCD.serverHostname" . }}
+    Admin Secret: {{ include "backingServices.argoCD.secretClusterName" . }}
+    Integreation Secret: {{ $argoCD.secretName }}
```
Roming22: Integreation -> Integration

charts/rhtap-backing-services/templates/_copy-scripts.tpl
```diff
@@ -0,0 +1,25 @@
+{{/*
+
+  POD container spec to copy scripts.
+
+*/}}
+{{- define "backingServices.copyScripts" -}}
+- name: copy-scripts
```
Roming22, line 7: Not related to this PR specifically, but I'm wondering if we should use init-containers in that pattern instead.

scripts/argocd-helper.sh
```diff
@@ -0,0 +1,123 @@
+#!/usr/bin/env bash
+#
+# Tests if the ArgoCD instance is available on the cluster by logging in.
+#
+# Uses the ArgoCD session, created by previsoly running "argocd login", to
```
Roming22: previsoly -> previously

scripts/argocd-helper.sh
```diff
@@ -0,0 +1,123 @@
...
+# generate a acccount token. The informaton is then stored on the ARGOCD_ENV_FILE
```
Roming22: a acccount -> an account
stored on -> stored in

scripts/argocd-helper.sh
```diff
@@ -0,0 +1,123 @@
...
+
+# Retries a few times until the ArgoCD instance is available.
+test_argocd_login() {
+    for i in {1..30}; do
+        echo "# [${i}/30] Testing ArgoCD login on '${ARGOCD_HOSTNAME}'..."
+        argocd_login &&
+            return 0
+
+        wait=$((i * 5))
+        echo "### [${i}/30] Waiting for ${wait} seconds before retrying..."
```
Roming22, line 57: We should not sleep on the 30th loop as we won't try logging in again.

scripts/argocd-store-token.sh
```diff
@@ -0,0 +1,89 @@
...
+# Waits for the environment file to be available.
+wait_for_env_file() {
+    echo "# Waiting for '${ARGOCD_ENV_FILE}' to be available..."
+    for i in {1..30}; do
+        [[ -r "${ARGOCD_ENV_FILE}" ]] &&
+            return 0
+
+        wait=$((i * 15))
+        echo "### [${i}/30] Waiting for '${ARGOCD_ENV_FILE}' to be available..."
+        sleep ${wait}
```
Roming22, line 33: See previous comment about extra sleep.

scripts/argocd-store-token.sh
```diff
@@ -0,0 +1,89 @@
...
+}
+
+# Waits for the environment file to be available.
+wait_for_env_file() {
+    echo "# Waiting for '${ARGOCD_ENV_FILE}' to be available..."
+    for i in {1..30}; do
+        [[ -r "${ARGOCD_ENV_FILE}" ]] &&
+            return 0
+
+        wait=$((i * 15))
```
Roming22, line 31: The step feels extra large. If my calculations are correct the max sleep time is 6975s, which seems overkill.

scripts/argocd-store-token.sh
```diff
@@ -0,0 +1,89 @@
...
+        fail "ARGOCD_ENV_FILE is not set!"
+
+    wait_for_env_file ||
+        fail "ARGOCD_ENV_FILE='${ARGOCD_ENV_FILE}' not found or not readable!"
+
+    echo "# Sourcing ArgoCD credentials from '${ARGOCD_ENV_FILE}'..."
+    # shellcheck source=/dev/null
+    . "${ARGOCD_ENV_FILE}"
+
+    [[ -z "${ARGOCD_HOSTNAME}" ]] &&
```
Roming22, line 56: Why not use `--from-file` and skip the validation?

charts/rhtap-backing-services/templates/_copy-scripts.tpl
otaviof: That makes sense, +1

charts/rhtap-backing-services/templates/NOTES.txt
otaviof: Just fixed the typo, thanks!

scripts/argocd-helper.sh
otaviof: Thanks, fixed.

scripts/argocd-helper.sh
otaviof: Also fixed 🙏

scripts/argocd-helper.sh
otaviof: That's a fix we need to tackle on basically all scripts, good point. Could we tackle this in a single PR, please?

scripts/argocd-store-token.sh
otaviof: Before it happens the Helm Chart installation times out will kick in, but it's definitely extra long.

scripts/argocd-store-token.sh
otaviof: Interesting suggestion, I'll check that further.

scripts/argocd-store-token.sh
otaviof: Another fix/chore I'd like to tackle on a subsequent PR, please.

PR #44, opened 2024-06-24
RHTAPINST-49: OpenShift Pipelines Configuration

otaviof:
Moving the Tekton Chains configuration to the infrastructure chart, replacing the pre-deploy hook with a post-deploy job that will create the Tekton Chains resources.

Improved the Helm chart testing to wait for OpenShift Pipelines to be ready, and using official Chainguard images to run `cosign.`

After #43 and #42 🙏


charts/rhtap-infrastructure/templates/openshift-pipelines/job-tekton-chains.yaml
```diff
@@ -0,0 +1,54 @@
...
+  {{- $passwordSecretName := required
+        ".infrastructure.openShiftPipelines.tektonChains.passwordSecretName"
+        $osp.tektonChains.passwordSecretName
+  -}}
+  {{- $signingSecretName := required
+        ".infrastructure.openShiftPipelines.tektonChains.signingSecretName"
+        $osp.tektonChains.signingSecretName
+  -}}
+  {{- $secretExists := lookup "v1" "Secret" $osp.namespace $signingSecretName -}}
+  {{- if not $secretExists }}
```
Roming22: Secret will exist by default but will be empty, so you need to test for both existence and content.

charts/rhtap-infrastructure/templates/openshift-pipelines/job-tekton-chains.yaml
```diff
@@ -0,0 +1,54 @@
...
+      serviceAccountName: {{ printf "patch-%s" $osp.name }}
+      restartPolicy: Never
+      containers:
+        - name: tekton-chains-cosign
+          image: ghcr.io/sigstore/cosign/cosign:latest
+          env:
+            - name: COSIGN_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: {{ $passwordSecretName }}
```
Roming22: I think we can just generate the value for the user.

charts/rhtap-infrastructure/templates/openshift-pipelines/job-tekton-config.yaml
```diff
@@ -0,0 +1,82 @@
...
+  {{- if or
+        $osp.patchClusterTektonConfig.annotations
+        $osp.patchClusterTektonConfig.labels
+  }}
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  annotations:
+    helm.sh/hook: post-install,post-upgrade
```
Roming22, line 12: If this is run post-install, how is charts/rhtap-backing-services/templates/openshift-pipelines/tektonconfig.yaml applied?

charts/rhtap-infrastructure/templates/openshift-pipelines/tekton-chains-password.yaml
```diff
@@ -0,0 +1,28 @@
+{{- if .Values.infrastructure.openShiftPipelines.enabled }}
```
Roming22: This resource feels superfluous.

charts/rhtap-infrastructure/values.yaml
```diff
@@ -165,3 +171,29 @@ infrastructure:
...
+    # Patches the global TektonConfig resource.
+    patchClusterTektonConfig:
+      # Resource name to patch.
+      resourceName: config
+      # Annotations to add to the resource.
+      annotations: {}
+      # Labels to add to the resource.
+      labels: {}
+    # Tekton Chains settings.
+    tektonChains:
```
Roming22, line 193: This section can go away, I don't see it's value.

charts/rhtap-infrastructure/templates/openshift-pipelines/job-tekton-chains.yaml
otaviof: Okay, I'll check if the attribute `cosign.key` is present.

charts/rhtap-infrastructure/templates/openshift-pipelines/job-tekton-chains.yaml
otaviof: Indeed, and later on `cosign generate-key-pair` saves the password on Secret aforementioned. I'll adjust this too. Thanks.

charts/rhtap-infrastructure/templates/openshift-pipelines/tekton-chains-password.yaml
otaviof: That's the link with [RHDH toggle](https://github.com/otaviof/rhtap-cli/blob/9fb0557a67891d058ce7c9342db9eb58b1ac5b0b/charts/values.yaml.tpl#L107-L108). So, when RHDH is disabled we can also disable OpenShift Pipelines changes.

charts/rhtap-infrastructure/values.yaml
otaviof: That section controls Tekton Chains related settings, and replaces hook scripts. In fact, is quite relevant.

charts/rhtap-infrastructure/templates/openshift-pipelines/job-tekton-config.yaml
otaviof: The `rhtap-infrastructure` chart is installed before `rhtap-backing-services`, so it patches the `TektonConfig` the latter will modify. By doing this we replace another hook script.

Roming22: My bad, I thought it was in `rhtap-backing-services`.

charts/rhtap-infrastructure/values.yaml
Roming22: I don't think any of these values are controlled by the user.

charts/values.yaml.tpl
```diff
@@ -104,6 +104,17 @@ infrastructure:
...
       namespace: {{ $tpa.Namespace }}
+  openShiftPipelines:
+    enabled: {{ $rhdh.Enabled }}
+    patchClusterTektonConfig:
+      annotations:
+        meta.helm.sh/release-name: rhtap-backing-services
+        meta.helm.sh/release-namespace: {{ $argoCDNamespace }}
+      labels:
+        app.kubernetes.io/managed-by: Helm
+    tektonChains:
```
Roming22: Let's discuss tomorrow why that's needed.

Roming22 (commented):
Please rebase to simplify the review.

charts/rhtap-infrastructure/values.yaml
otaviof: These attributes have defaults, they should be good enough to work with the current version of OpenShift Pipelines, and allow us to modify as we move forward. The entries added to the `values.yaml` are, in fact, variables that we use in multiple resources to avoid duplication, among other reasons to define them.

Additionally, by having these attributes we can leverage their relationship via `values.yaml.tpl`.

charts/values.yaml.tpl
otaviof: It's needed to replace this:

https://github.com/redhat-appstudio/rhtap-cli/blob/2cff2c50dd15a083c7f555cfd416d6f47e2d657c/charts/rhtap-backing-services/hooks/pre-deploy.sh#L6-L17

Thus we can run these type of interactions in the cluster, without the need for extra local tools to run `rhtap-cli`.

charts/rhtap-infrastructure/templates/openshift-pipelines/job-tekton-config.yaml
otaviof: Sure 👍

otaviof:
> Please rebase to simplify the review.

Sure! Just rebased, please consider.

charts/rhtap-infrastructure/values.yaml
otaviof: For the record, I moved the defaults to the `values.yaml` on `rhtap-infrastructure` instead of setting them up on `values.yaml.tpl`.

Roming22 (approved):

PR #47, opened 2024-06-25
RHTAPINST-51: Feature Properties Dictionary

otaviof:
Allows to define additional properties per feature, initially for the `catalogURL` reference.

### DIFF
--- a/charts/values.yaml.tpl
+++ b/charts/values.yaml.tpl
@@ -151,6 +151,10 @@ backingServices:
     secretName: rhtap-argocd-integration
     ingressDomain: {{ $ingressDomain }}
 
+#
+# rhtap-integrations
+#
+
 # integrations:
 #   acs:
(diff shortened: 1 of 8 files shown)


charts/values.yaml.tpl
```diff
@@ -151,35 +151,23 @@ backingServices:
     secretName: rhtap-argocd-integration
     ingressDomain: {{ $ingressDomain }}
 
-# integrations:
```
Roming22, line 154: Please keep the `integrations`  for the moment. You can remove the `developerHub` and `catalogUrl` fields though.

We can remove it once we've exposed everything through `config.yaml`.

otaviof: Sure, I've reverted this chunk. Please consider.

Roming22 (approved):

PR #75, opened 2024-07-02
RHTAPINST-61: Generate a StackRox API Token

otaviof:
When ACS is installed an API token is generated and stored as a secret.

### DIFF
--- a/charts/rhtap-backing-services/templates/openshift-acs/job-stackrox-api.yaml
+++ b/charts/rhtap-backing-services/templates/openshift-acs/job-stackrox-api.yaml
@@ -1,0 +1,58 @@
+{{- if .Values.backingServices.acs.enabled -}}
+  {{- $acs := .Values.backingServices.acs -}}
+  {{- $name := printf "%s-post-deploy-%d" $acs.name .Release.Revision }}
+#
+# Generates a token for StackRox API.
+#
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  annotations:
+    helm.sh/hook: post-install,post-upgrade
+    helm.sh/hook-delete-policy: hook-succeeded
+  labels:
+    {{- include "rhtap-backing-services.labels" . | nindent 4 }}
+  namespace: {{ $acs.namespace }}
+  name: {{ $name }}
+spec:
+  template:
+    spec:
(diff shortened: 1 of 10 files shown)


charts/rhtap-backing-services/templates/openshift-acs/job-stackrox-api.yaml
```diff
@@ -0,0 +1,58 @@
+{{- if .Values.backingServices.acs.enabled -}}
+  {{- $acs := .Values.backingServices.acs -}}
+  {{- $name := printf "%s-post-deploy-%d" $acs.name .Release.Revision }}
```
Roming22, line 3: Why use of the revision #?

Roming22:
I just did a test and the acs tasks of the software template failed. My guess is that something is not right during deployment as I don't have a `rhtap-acs-integration` secret.

charts/rhtap-backing-services/templates/openshift-acs/job-stackrox-api.yaml
```diff
@@ -0,0 +1,58 @@
...
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  annotations:
+    helm.sh/hook: post-install,post-upgrade
+    helm.sh/hook-delete-policy: hook-succeeded
+  labels:
+    {{- include "rhtap-backing-services.labels" . | nindent 4 }}
+  namespace: {{ $acs.namespace }}
```
Roming22, line 16: That job has to be run in the rhtap namespace as the `rhtap-acs-integration` is expected to be there.

charts/rhtap-backing-services/templates/openshift-acs/job-stackrox-api.yaml
otaviof: Let's tackle the revision numbers, and the Helm labels to delete/prune jobs in a another PR, as we've spoken.

charts/rhtap-backing-services/templates/openshift-acs/job-stackrox-api.yaml
otaviof: Cool, thanks.

Roming22 (approved):

PR #78, opened 2024-07-04
RHTAPINST-36: Trusted Profile Analyzer v0.1.0

otaviof:
Updated `openshift-helm-charts/redhat-trusted-profile-analyzer` to version v0.1.0. additionally, allowing ACS to take longer to start up.


charts/rhtap-tpa/Chart.yaml
```diff
@@ -3,9 +3,9 @@ apiVersion: v2
 name: rhtap-tpa
 description: RHTAP Trusted Profile Analyzer
 type: application
-version: 0.0.1
+version: 0.1.0
```
Roming22, line 6: I think this value should track the TPA version being installed rather than the TPA chart version.

charts/rhtap-tpa/Chart.yaml
```diff
@@ -3,9 +3,9 @@ apiVersion: v2
...
 
 dependencies:
-  - name: trustification
-    version: 0.2.0
+  - name: redhat-trusted-profile-analyzer
+    version: 0.1.0
```
Roming22, line 10: Can we use the latest chart (0.4.0 if I remember correctly)?

Roming22 (commented):
I would also change the commit message to reference the TPA version rather than the TPA chart version.

charts/rhtap-tpa/Chart.yaml
Roming22: I got that wrong, you have the right version.

charts/rhtap-tpa/Chart.yaml
Roming22: I'll merge as is, we can discuss later what makes more sense.

Roming22 (approved):

PR #85, opened 2024-07-09
fix: github app integration

Roming22:


vendor/github.com/google/go-github/scrape/apps.go
```diff
@@ -115,6 +115,8 @@ type AppManifest struct {
 	Name *string `json:"name,omitempty"`
 	//Required. The homepage of your GitHub App.
 	URL *string `json:"url,omitempty"`
+	//Required. The endpoint for authentication.
+	CallbackURL *string `json:"callback_url,omitempty"`
```
otaviof: That's something we can only change via `go.mod`.

Roming22: https://github.com/google/go-github/issues/3203

PR #87, opened 2024-07-10
Migrate RHTAP installer CI to RHTAP cli

xinredhat:


.tekton/integration-tests/tasks/rhtap-install.yaml
```diff
@@ -35,5 +42,46 @@ spec:
...
+          gitlab:
+            token: "${GITLAB__TOKEN}"
+          quay:
+            dockerconfigjson: |
+        $(echo "${QUAY__DOCKERCONFIGJSON}" | sed 's/^/      /')
+            token: "${QUAY__API_TOKEN}"
+        EOF
+        )
+
+        # Use awk to replace the integrations section
```
ascerra: I was looking at the test output of the last run and it doesn't look like these replaced values in the '# integrations:' section were applied

Link to most recent run as of today: https://console.redhat.com/preview/application-pipeline/workspaces/rhtap-shared/applications/rhtap-cli/pipelineruns/rhtap-cli-s5m25-pp8mf/logs?task=rhtap-install

If you search for "# integrations" you'll see that the values are the same as in the repo with the values being "".

.tekton/integration-tests/tasks/rhtap-install.yaml
```diff
@@ -35,5 +42,46 @@ spec:
...
+          found && /^# rhtap-dh/ { found = 0 }
+          !found { print }
+        ' "$tpl_file" > tmpfile && mv tmpfile "$tpl_file"
+
+        echo "make build"
         make build
-        ./bin/rhtap-cli deploy --config ./config.yaml --kube-config "$KUBECONFIG" --debug
+
+        echo "install"
+        ./bin/rhtap-cli deploy --config ./config.yaml --kube-config "$KUBECONFIG" --debug --log-level=debug
```
ascerra: I don't believe we should be using --debug --log-level=debug since we are attempting to add secrets to the config. This will essentially just print our secrets in plain text in the logs.

xinredhat: i see, I am debugging this CI by using it

xinredhat: Done

integration-tests/tasks/rhtap-install.yaml
```diff
@@ -35,5 +45,37 @@ spec:
...
+        # Export the environment variables from the secret
+        cat /workspace/config.env > config.sh
+        chmod +x config.sh
+        source ./config.sh
+
+        # Path to your values.yaml.tpl file
+        tpl_file="charts/values.yaml.tpl"
+
+        # Create the new integrations section
+        cat <<EOF >> "$tpl_file"
```
rhopp, line 57: Is this the supported way? I thought this should be done by calling various `rhtap-cli integration *` commands and there shouldn't be a reason to touch this template.

xinredhat: >Is this the supported way?

it's not an official supported way. For now rhtap-cli cannot work with an existing GithubApp, which means command `rhtap-cli integration github-app` only work for `create`, `uddate` doesn't work. and The `create` process is not fully automatic.

xinredhat: And I think we need 2 CI for installation verification, one is full install, one is min install by using external ACS, Quay, etc

xinredhat:
the failure is expected.

integration-tests/tasks/rhtap-install.yaml
rhopp: Yeah... originally I didn't realize that. Thanks

PR #102, opened 2024-07-17
RHTAPINST-81: Quay Integration

otaviof:
Using a script to initialise Quay super user and create a docker-registry secret for the RHTAP integration.

PR #104, opened 2024-07-17
feat: add 'integration quay' subcommand

Roming22:


pkg/integrations/quay.go
```diff
@@ -0,0 +1,182 @@
...
+	token            string // API token credentials
+	url              string // quay URL
+}
+
+// PersistentFlags sets the persistent flags for the Quay integration.
+func (q *QuayIntegration) PersistentFlags(p *pflag.FlagSet) {
+	p.BoolVar(&q.force, "force", q.force,
+		"Overwrite the existing secret")
+
+	p.StringVar(&q.dockerconfigjson, "dockerconfigjson", q.dockerconfigjson,
```
otaviof, line 37: I think we should follow the same pattern than `oc create secret`, which in this case instead of `--dockerconfigjson` we would adopt the known flag `--from-file`.

pkg/integrations/quay.go
```diff
@@ -0,0 +1,182 @@
...
+
+// secretName returns the secret name for the integration. The name is "lazy"
+// generated to make sure configuration is already loaded.
+func (q *QuayIntegration) secretName() types.NamespacedName {
+	feature, _ := q.cfg.GetFeature(config.RedHatDeveloperHub)
+	return types.NamespacedName{
+		Namespace: feature.GetNamespace(),
+		Name:      "rhtap-quay-integration",
+	}
+}
```
otaviof, line 100: For the record, later on we should export this type of routine to a common place. There's a couple examples this and ensure the installer namespace is created, etc.

pkg/integrations/quay.go
```diff
@@ -0,0 +1,182 @@
...
+			Namespace: q.secretName().Namespace,
+			Name:      q.secretName().Name,
+		},
+		Type: corev1.SecretTypeOpaque,
+		Data: map[string][]byte{
+			"dockerconfigjson": []byte(q.dockerconfigjson),
+			"token":            []byte(q.token),
+			"url":              []byte(q.url),
+		},
+	}
```
otaviof: We should follow common Kubernetes convention to have a secret typed `docker-registry`, the only adaptation needed here is storing the `--from-file` argument as `.dockerconfigjson` attribute. 🙏

otaviof (commented):
Cool, great work 💪 Let's talk about the remarks here in light of #102, please.

pkg/integrations/quay.go
Roming22: Agreed.

pkg/integrations/quay.go
Roming22: Updated to use the requested secret type. The other comment will track the use of `--from-file`.

pkg/integrations/quay.go
Roming22: Can we deal with all the usecases to allow a user to specify the secret values as a later improvment?

otaviof: Deal 👍

PR #121, opened 2024-07-25
RHTAPINST-97: Container Image Build

otaviof:
Builds a container image out of this project. Two new `Makefile` targets have been added:

```bash
make image
make image-run
```


otaviof:
@flacatus, as we've spoken, let's merge this 🙏

PR #142, opened 2024-08-01
fix: use init containers to copy scripts

Roming22:
This solves a possible race condition where a script could be executed before it has been written to disk.


otaviof (approved):
Nice one, @Roming22!

PR #149, opened 2024-08-07
RHTAPINST-108: Structured Helm Release Monitoring

otaviof:
Inspect Helm release payload to select `ProjectRequest` resources, which triggers monitoring the creation of the respective Kubernetes namespace. This approach replaces custom scripting for a structured approach.


pkg/monitor/monitor.go
```diff
@@ -0,0 +1,101 @@
...
+
+			// Run the monitor function, if successful remove it from the queue,
+			// or retries after grace period.
+			if err := m.queue[i](); err == nil {
+				logger.Debug("Monitor function succeeded!")
+				m.queue = append(m.queue[:i], m.queue[i+1:]...)
+			} else {
+				logger.Debug("Monitor function failed, retrying...")
+				i++
+				time.Sleep(300 * time.Millisecond)
```
Roming22: What's the purpose of this retry? There's a global one line 90.

pkg/monitor/monitor.go
```diff
@@ -0,0 +1,101 @@
...
+	return nil
+}
+
+// Watch waits for all monitoring functions to complete, or until the timeout is
+// reached. Returns error if the queue is not empty after timeout.
+func (m *Monitor) Watch(timeout time.Duration) error {
+	start := time.Now()
+	logger := m.logger.With("timeout", timeout.String(), "start", start.String())
+	for {
+		for i := 0; i < len(m.queue); {
```
Roming22: I don't understand the value of this nested loop.

pkg/monitor/monitor.go
otaviof: Indeed, this retry has no purpose, in fact, the actual retry of these failed functions will happen in the subsequent parent loop interaction. I refactored this part, please consider.

pkg/monitor/monitor.go
otaviof: The nested loop is need for the retry mechanism, the failed functions remain on the `m.queue`, thus the parent loop is controlling retries of failed `m.queue` items.

Roming22: What I mean is that the following code seems to lead to the same result while being easier to read.

```
	for len(m.queue) > 0 {
		if time.Since(start) >= timeout {
			return errors.New("timeout reached")
		}

		if err := m.queue[0](); err == nil {
			logger.Debug("Monitor function succeeded!")
			m.queue = m.queue[1:]
		} else {
			logger.Debug("Monitor function failed!")
			time.Sleep(300 * time.Millisecond)	
		}
	}
	logger.Debug("Monitoring complete, queue is empty!")
	return nil
```

Roming22: I would also bump the delay to 2s.

otaviof: Okay, sure! Good suggestions here, please consider latest changes.

Roming22 (approved):

PR #158, opened 2024-08-08
fix: sanitize GH App URL

Roming22:


pkg/githubapp/githubapp.go
```diff
@@ -113,7 +114,12 @@ func (g *GitHubApp) oAuth2Workflow(
 				oAppConfigCh <- result
 				return
 			}
-			fmt.Fprintf(w, gitHubAppSuccessfullyCreatedTmpl, *result.appConfig.HTMLURL)
+
+			// Sanitize URL
+			u, _ := url.Parse(*result.appConfig.HTMLURL)
+			b, _ := u.MarshalBinary()
```
otaviof, line 120: Should we capture and `return err` these two errors?

Roming22: 1. I would not expect the url to be problematic in anyway. We're just making snyk happy.
2. The app has been created, so we want to return successfully to ensure the secret is created.

otaviof (approved):

PR #177, opened 2024-08-20
fix: DH GitLab config

Roming22:
- [RHTAPBUGS-1300](https://issues.redhat.com//browse/RHTAPBUGS-1300)
- Remove duplicates

### DIFF
--- a/pkg/integrations/gitlab.go
+++ b/pkg/integrations/gitlab.go
@@ -25,10 +25,8 @@ type GitLabIntegration struct {
 
 	force bool // overwrite the existing secret
 
-	host         string // GitLab host
-	clientId     string // GitLab application client id
-	clientSecret string // GitLab application client secret
-	token        string // API token credentials
+	host  string // GitLab host
+	token string // API token credentials
 }
 
 // PersistentFlags sets the persistent flags for the GitLab integration.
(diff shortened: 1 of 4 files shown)


pkg/integrations/gitlab.go
```diff
@@ -38,10 +38,10 @@ func (g *GitLabIntegration) PersistentFlags(p *pflag.FlagSet) {
...
 	p.StringVar(&g.host, "host", g.host,
 		"GitLab host, defaults to 'gitlab.com'")
-	p.StringVar(&g.clientId, "app-id", g.clientId,
-		"GitLab application client id")
-	p.StringVar(&g.clientSecret, "app-secret", g.clientSecret,
-		"GitLab application client secret")
+	// p.StringVar(&g.clientId, "app-id", g.clientId,
+	// 	"GitLab application client id")
+	// p.StringVar(&g.clientSecret, "app-secret", g.clientSecret,
+	// 	"GitLab application client secret")
```
otaviof: Please remove these lines instead of just comment them out.

otaviof (changes requested):
Just a minor change, please.

pkg/integrations/gitlab.go
Roming22: Done

otaviof (changes requested):
By the way, could you also unwire the attributes here: 

https://github.com/redhat-appstudio/rhtap-cli/blob/70bd7e239842e095a473a2b37c9c50657bc4d19a/pkg/integrations/gitlab.go#L29-L30

Roming22:
@otaviof I've now fully removed the feature.

PR #182, opened 2024-08-21
fix: install with a different namespace prefix

Roming22:
[RHTAPBUGS-1302](https://issues.redhat.com//browse/RHTAPBUGS-1302)
[RHTAPBUGS-1303](https://issues.redhat.com//browse/RHTAPBUGS-1303)


Roming22:
Here's the output of a manual test. Note that the software templates repo must be forked and modified to reference the new namespace.

image

otaviof (approved):

PR #196, opened 2024-09-05
fix: fail install on missing csr-signer secret

Roming22:


otaviof (approved):
Just a `nit` comment, but looks good. Thanks!

hack/fix-csr-signer.sh
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/env bash
+
```
otaviof: Would be good to have a comment header explaining why this scripts exists.

Roming22: I'm having issues with the workaround. Since it's not critical for the release, I removed it from this PR.

Roming22:
#199 is the right fix

PR #207, opened 2024-09-10
fix: get ocp clients from a trusted source

Roming22:


Roming22:
@flacatus It is my understanding that the failure is unrelated to the change. If you and @otaviof agree with this change, please force merge.

Once it's cherry-picked to the release branch, please do the same.

flacatus:
the failure is not related with this PR

otaviof (dismissed):
Instead of downloading external software, let's try to use `quay.io/codeready-toolchain/oc-client-base:latest` as the base for the `rhtap-cli` image?

Hook scripts won't work well on Windows, in other words, are not portable, and installing external things on the user's workstation is something we must avoid.

Additionally, let's update documentation to show `kubectl` must be installed.

flacatus:
>Instead of downloading external software, let's try to use quay.io/codeready-toolchain/oc-client-base:latest as the base for the rhtap-cli image?

If we do this the konflux will detect codeready cves and willl not release

Roming22:
@otaviof we can redirect windows users to use the container image, and the script cleans up after itself, removing the binaries. I would think that's good enough for the moment.

otaviof:
> > Instead of downloading external software, let's try to use quay.io/codeready-toolchain/oc-client-base:latest as the base for the rhtap-cli image?
> 
> If we do this the konflux will detect codeready cves and willl not release

By downloading a random binary from the internet during runtime we lose visibility from any CVEs.

otaviof:
So this PR is reducing the portability and introducing a unwanted binary from the internet which won't be subject to Snyk checks.

Roming22:
@otavio, I agree, but that's the only path forward we have at the moment to deliver through Konflux.

Roming22:
Also, the binary is produced by RH, is downloaded from a RH controlled URL, and is expected to be production worthy. Therefore the solution was considered as acceptable during the program call.

Once a better and Konflux compliant solution is available, we should use it, as your concerns are valid.

PR #209, opened 2024-09-11
feat: Get oc cli from redhat registry

flacatus:


flacatus:
Do not review, just testing

PR #213, opened 2024-09-11
feat: Add labels to cli image to avoid ec violations when we release

flacatus:


flacatus:
``` log
✕ [Violation] labels.disallowed_inherited_labels
  ImageRef: quay.io/rhtap/rhtap-cli@sha256:f91bf637e341196fb5a8256ff673ed2de2f46eb813719d67fd0ff6570c875f74
  Reason: The "summary" label should not be inherited from the parent image
  Title: Disallowed inherited labels
  Description: Check that certain labels on the image have different values than the labels from the parent image. If the label is
  inherited from the parent image but not redefined for the image, it will contain an incorrect value for the image. Use the rule
  data `disallowed_inherited_labels` key to set the list of labels to check, or the `fbc_disallowed_inherited_labels` key for fbc
  images. To exclude this rule add "labels.disallowed_inherited_labels:summary" to the `exclude` section of the policy
  configuration.
  Solution: Update the image build process to overwrite the inherited labels.
```

Roming22 (approved):
/lgtm

Roming22 (approved):
/lgtm

PR #222, opened 2024-09-12
Sync 1.2 with main

Roming22:


Roming22:
1.2.1 cancelled as we don't have the bandwidth to manage the release

PR #244, opened 2024-09-26
RHTAPINST-149: Wait for Operators Rollout

otaviof:
All OLM managed operator have a "olm.managed=true" label, so we can use this to wait for the rollout of the operator to complete.


installer/config.yaml
Roming22, line 1: You might want to revert those changes

otaviof: Indeed, please consider latest changes.

PR #247, opened 2024-09-30
RHTAPINST-137: Nexus integration with rhtap

prietyc123:
Integrating nexus as container registry for rhtap

### DIFF
--- a/installer/charts/rhtap-dh/hooks/post-deploy.sh
+++ b/installer/charts/rhtap-dh/hooks/post-deploy.sh
@@ -17,26 +17,40 @@ get_binaries() {
     exit 1
 }
 
+patch_serviceaccount() {
+    local NAMESPACE="$1"
+    local SA="$2"
+
+    echo -n "- Patching ServiceAccount '$SA' in '$NAMESPACE': "
+
+    # Wait until the ServiceAccount is available
+    until "$KUBECTL" get serviceaccounts --namespace "$NAMESPACE" "$SA" >/dev/null 2>&1; do
+        echo -n "_"
+        sleep 2
+    done
+
+    # Check for quay-auth and nexus-auth secrets and patch if present
+    for SECRET_NAME in nexus-auth quay-auth; do
+        SECRET=$("$KUBECTL" get secret "$SECRET_NAME" --namespace "$NAMESPACE" --ignore-not-found)
+        if [ -n "$SECRET" ]; then
+            echo -n "."
+            "$KUBECTL" patch serviceaccounts --namespace "$NAMESPACE" "$SA" --patch "
+secrets:
+    - name: $SECRET_NAME
+imagePullSecrets:
+    - name: $SECRET_NAME
+" >/dev/null
+            echo "OK"
+        fi
+    done
+}
+
 app_namespaces() {
-    ### Workaround
-    ### Helm does not support the patching of resources.
     NAMESPACE="$INSTALLER__DEVELOPERHUB__NAMESPACE"
 
     for env in "development" "prod" "stage"; do
         for SA in "default" "pipeline"; do
-            echo -n "- Patching ServiceAccount '$SA' in '$NAMESPACE-app-$env': "
-            until "$KUBECTL" get serviceaccounts --namespace "$NAMESPACE-app-$env" "$SA" >/dev/null 2>&1; do
-                echo -n "_"
-                sleep 2
-            done
-            echo -n "."
-            "$KUBECTL" patch serviceaccounts --namespace "$NAMESPACE-app-$env" "$SA" --patch "
-secrets:
-    - name: quay-auth
-imagePullSecrets:
-    - name: quay-auth
-" >/dev/null
-            echo "OK"
+            patch_serviceaccount "$NAMESPACE-app-$env" "$SA"
         done
     done
 }
--- a/installer/charts/rhtap-dh/templates/app-config-content.yaml
+++ b/installer/charts/rhtap-dh/templates/app-config-content.yaml
@@ -2,8 +2,16 @@
 app:
   title: Red Hat Developer Hub
   baseUrl: ${BACKEND_URL}
+# Lookup for all the required secrets
 {{- $argocdSecretObj := (lookup "v1" "Secret" .Release.Namespace "rhtap-argocd-integration") }}
 {{- $argocdSecretData := ($argocdSecretObj.data | default dict) }}
+{{- $ghSecretObj := (lookup "v1" "Secret" .Release.Namespace "rhtap-github-integration") }}
+{{- $gitlabSecretObj := (lookup "v1" "Secret" .Release.Namespace "rhtap-gitlab-integration") }}
+{{- $jenkinsSecretObj := (lookup "v1" "Secret" .Release.Namespace "rhtap-jenkins-integration") }}
+{{- $quaySecretObj := (lookup "v1" "Secret" .Release.Namespace "rhtap-quay-integration") }}
+{{- $quaySecretData := ($quaySecretObj.data | default dict) }}
+{{- $nexusSecretObj := (lookup "v1" "Secret" .Release.Namespace "rhtap-nexus-integration") }}
+
 {{- if $argocdSecretData }}
 argocd:
   appLocatorMethods:
@@ -19,7 +27,6 @@ argocd:
 auth:
   environment: production
   providers:
-  {{- $ghSecretObj := (lookup "v1" "Secret" .Release.Namespace "rhtap-github-integration") }}
   {{- if $ghSecretObj }}
     github:
       production:
@@ -52,7 +59,7 @@ catalog:
       - Location
       - Template
       - API
-{{- if (lookup "v1" "Secret" .Release.Namespace "rhtap-jenkins-integration") }}
+{{- if $jenkinsSecretObj }}
 jenkins:
   instances:
     - name: default
@@ -61,7 +68,7 @@ jenkins:
       apiKey: ${JENKINS__TOKEN}
 {{- end }}
 integrations:
-{{- if (lookup "v1" "Secret" .Release.Namespace "rhtap-github-integration") }}
+{{- if $ghSecretObj }}
   github:
     - host: ${GITHUB__HOST}
       apps:
@@ -72,7 +79,7 @@ integrations:
           webhookSecret: ${GITHUB__APP__WEBHOOK__SECRET}
           privateKey: ${GITHUB__APP__PRIVATE_KEY}
 {{- end }}
-{{- if (lookup "v1" "Secret" .Release.Namespace "rhtap-gitlab-integration") }}
+{{- if $gitlabSecretObj }}
   gitlab:
     - host: ${GITLAB__HOST}
       apiBaseUrl: https://${GITLAB__HOST}/api/v4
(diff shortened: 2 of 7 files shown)


installer/charts/rhtap-dh/templates/extra-env.yaml
```diff
@@ -63,3 +63,9 @@ data:
     QUAY__API_TOKEN: {{ $quaySecretData.token }}
 {{- end }}
     QUAY__URL: {{ $quaySecretData.url }}
```
Roming22, line 66: You need to surround all of the QUAY vars (L62-65) with
```
{{- if $quaySecretData }}
{{- end }}

File: installer/charts/rhtap-dh/hooks/post-deploy.sh [hunk 2]
```diff
@@ -17,26 +17,56 @@ get_binaries() {
...
+    echo -n "- Patching ServiceAccount '$SA' in '$NAMESPACE': "
+
+    # Wait until the ServiceAccount is available
+    until "$KUBECTL" get serviceaccounts --namespace "$NAMESPACE" "$SA" >/dev/null 2>&1; do
+        echo -n "_"
+        sleep 2
+    done
+    echo -n "."
+
+    # Check for quay-auth and nexus-auth secrets and patch if present
```
Roming22, line 32: You should be able to simplify (untested):
```
    # Check for quay-auth and nexus-auth secrets and patch if present
    for SECRET_NAME in nexus-auth quay-auth; do
        SECRET=$("$KUBECTL" get secret --namespace "$NAMESPACE" --ignore-not-found)
        if [ -n "$SECRET" ]; then
            "$KUBECTL" patch serviceaccounts --namespace "$NAMESPACE" "$SA" --patch "
secrets:
    - name: $SECRET_NAME
imagePullSecrets:
    - name: $SECRET_NAME
imagePullSecrets:
" >/dev/null
        fi
        echo "OK"
    done
```

installer/charts/rhtap-dh/templates/app-config-content.yaml
```diff
@@ -90,11 +90,22 @@ proxy:
         Authorization: 'Bearer ${QUAY__API_TOKEN}'
       {{- end }}
       changeOrigin: true
-      # Change to "false" in case of using self hosted quay instance with a self-signed certificate
+      secure: true
+    '/nexus/api':
```
Roming22, line 103: You need to surround all of the quay/nexus config with
```
{{- if $****Secret}}
{{- end }
```
Same on lines 105-108

installer/charts/rhtap-dh/templates/extra-env.yaml
```diff
@@ -63,3 +63,9 @@ data:
...
+{{- $nexusSecretObj := (lookup "v1" "Secret" .Release.Namespace "rhtap-nexus-integration") }}
+{{- $nexusSecretData := ($nexusSecretObj.data | default dict) }}
+{{- if $nexusSecretData }}
+    NEXUS__TOKEN: {{ $nexusSecretData.token }}
```
Roming22: Are there public repositories on Nexus? If so the token might be optional. You would need to update `nexus.go` too.

installer/charts/values.yaml.tpl
```diff
@@ -6,9 +6,10 @@
 {{- $gitops := required "GitOps settings" .Installer.Features.openShiftGitOps -}}
 {{- $pipelines := required "Pipelines settings" .Installer.Features.openShiftPipelines -}}
 {{- $quay := required "Quay settings" .Installer.Features.redHatQuay -}}
+{{- $quayEnabled := not (lookup "v1" "Secret" .Installer.Namespace "rhtap-nexus-integration") -}}
```
Roming22: That should be `nexusEnabled`. You need to add logic to support both `quay.Enabled` and/or `nexusEnabled`.

installer/charts/values.yaml.tpl
```diff
@@ -6,9 +6,10 @@
...
 {{- $rhdh := required "RHDH settings" .Installer.Features.redHatDeveloperHub -}}
 {{- $ingressDomain := required "OpenShift ingress domain" .OpenShift.Ingress.Domain -}}
-{{- $minIOOperatorEnabled := or $tpa.Enabled $quay.Enabled -}}
+{{- $minIOOperatorEnabled := or $tpa.Enabled $quayEnabled -}}
```
Roming22: This change looks wrong. Why do we need to install minIO if TPA is disabled and an external Nexus instance is used?

installer/charts/values.yaml.tpl
```diff
@@ -27,7 +28,7 @@ openshift:
     - rhacs-operator
     - {{ $acs.Namespace }}
 {{- end }}
-{{- if $quay.Enabled }}
+{{- if $quayEnabled }}
```
Roming22: All those changes from `quay.Enabled` to `quayEnabled` look wrong to me. Maybe I missed something?

Roming22 (changes requested):

installer/charts/rhtap-dh/hooks/post-deploy.sh
prietyc123: sure

installer/charts/rhtap-dh/templates/extra-env.yaml
prietyc123: I am also looking for this but so far couldn't find a way out.

installer/charts/values.yaml.tpl
prietyc123: IIUC we only need to worry about enable/disable quay but not nexus because we are not installing nexus operator. We only want the `rhtap-cli integrations` . Here with `quayEnabled`, I am disabling quay if `rhtap-nexus-integration` secrets are available but if not just enable quay to get it installed. Let me know if my understanding is not correct

installer/charts/values.yaml.tpl
prietyc123: Here as you see it does not referring to nexus but quay. As I explained in my above comment that if `nexus` is there we are disabling quay and hence minIO should not be functional.

installer/charts/values.yaml.tpl
prietyc123: ptal https://github.com/redhat-appstudio/rhtap-cli/pull/247#discussion_r1794723328

prietyc123:
@Roming22 I addressed your review comment in [commit](https://github.com/redhat-appstudio/rhtap-cli/pull/247/commits/6f17fe92e46a1ed7f1db73309950cae55cc52964) 
Could you please take a look. Also I have responded to your concern related to enable/disable quay

installer/charts/rhtap-dh/templates/extra-env.yaml
```diff
@@ -59,7 +59,12 @@ data:
 {{- end }}
 {{- $quaySecretObj := (lookup "v1" "Secret" .Release.Namespace "rhtap-quay-integration") }}
 {{- $quaySecretData := ($quaySecretObj.data | default dict) }}
-{{- if $quaySecretData.token }}
+{{- if $quaySecretData }}
```
Roming22, line 62: The token is optional (not needed when the image repo is public).
```
{{- if $quaySecretData }}
    {{- if $quaySecretData.token }}
    QUAY__API_TOKEN: {{ $quaySecretData.token }}
    {{- end }}
    QUAY__URL: {{ $quaySecretData.url }}
{{- end }}
```

installer/charts/values.yaml.tpl
Roming22: 1. You would be missing an 'or' condition I believe. `quayEnabled` will be true if there is no nexus integration even if the user disabled quay in the config.yaml.
2. The logic you are putting in place is saying "you cannot have quay and nexus at the same time". Do we know that it is true? Is DH breaking when that is the case?

installer/charts/values.yaml.tpl
Roming22: Let's figure out the logic in the other thread, and then see how it impacts this.

installer/charts/values.yaml.tpl
Roming22: Let's figure out the logic in the other thread, and then see how it impacts this.

Roming22 (changes requested):

prietyc123:
> The logic you are putting in place is saying "you cannot have quay and nexus at the same time". Do we know that it is true? > Is DH breaking when that is the case?

Ok I need to confirm on this. As per my understanding, I was under impression that we should not have quay and nexus at the same time

prietyc123:
@Roming22 I have verified we can have quay and nexus at the same time.

installer/charts/rhtap-dh/templates/extra-env.yaml
```diff
@@ -59,7 +59,14 @@ data:
...
+    {{- if $quaySecretData.token }}
     QUAY__API_TOKEN: {{ $quaySecretData.token }}
-{{- end }}
+    {{- end }}
     QUAY__URL: {{ $quaySecretData.url }}
+{{- end }}
+{{- $nexusSecretObj := (lookup "v1" "Secret" .Release.Namespace "rhtap-nexus-integration") }}
+{{- $nexusSecretData := ($nexusSecretObj.data | default dict) }}
+{{- if $nexusSecretData }}
+    NEXUS__URL: {{ $nexusSecretData.url }}
```
Roming22, line 71: I'm surprised there is no token for private image repositories. OK to merge as is until we figure it out.

installer/charts/values.yaml.tpl
```diff
@@ -6,9 +6,12 @@
 {{- $gitops := required "GitOps settings" .Installer.Features.openShiftGitOps -}}
 {{- $pipelines := required "Pipelines settings" .Installer.Features.openShiftPipelines -}}
 {{- $quay := required "Quay settings" .Installer.Features.redHatQuay -}}
+{{- $nexusSecret := (lookup "v1" "Secret" .Installer.Namespace "rhtap-nexus-integration") -}}
+{{- $quaySecret := (lookup "v1" "Secret" .Installer.Namespace "rhtap-quay-integration") -}}
+{{- $quayEnabled := not (or $nexusSecret $quaySecret) -}}
```
Roming22: 1. The variable name is confusing
2. If quay is deployed by the installer, the secret doesn't exist yet.

installer/charts/values.yaml.tpl
```diff
@@ -6,9 +6,12 @@
...
 {{- $rhdh := required "RHDH settings" .Installer.Features.redHatDeveloperHub -}}
 {{- $ingressDomain := required "OpenShift ingress domain" .OpenShift.Ingress.Domain -}}
-{{- $minIOOperatorEnabled := or $tpa.Enabled $quay.Enabled -}}
+{{- $minIOOperatorEnabled := or $tpa.Enabled $quayEnabled -}}
```
Roming22: That would install minio if there was Nexus and no TPA. Feels wrong. I don't think you need to change that condition, here or later in the file.

pkg/subcmd/integration_nexus.go
```diff
@@ -0,0 +1,98 @@
...
+
+	nexusIntegration *integrations.NexusIntegration // nexus integration
+
+	dockerconfigjson string // credentials to push/pull from the registry
+}
+
+var _ Interface = &IntegrationNexus{}
+
+const nexusIntegrationLongDesc = `
+Manages the nexus integration with RHTAP, by storing the required
```
Roming22: Nit pick: Use 'Nexus' when referring to the product name instead of 'nexus'.

pkg/subcmd/integration_nexus.go
```diff
@@ -0,0 +1,98 @@
...
+credentials required by the RHTAP services to interact with nexus.
+
+The credentials are stored in a Kubernetes Secret in the configured namespace
+for RHDH.
+
+
+The given dockerconfig must include the repository path.
```
Roming22: Remove that line. I believe it was for a quay issue.

Roming22 (changes requested):

installer/config.yaml
```diff
@@ -19,7 +19,7 @@ rhtapCLI:
       properties:
         catalogURL: https://github.com/redhat-appstudio/tssc-sample-templates/blob/release/all.yaml
     redHatAdvancedClusterSecurity:
-      enabled: *rhdhEnabled
+      enabled: false
```
Roming22: Please revert.

installer/charts/values.yaml.tpl
Roming22: Please revert

Roming22 (approved):
/lgtm

PR #248, opened 2024-09-30
fix: move CI install logic to a script

Roming22:
This allow the logic to be tested in PRs, as well as troubleshooting the CI installation process in local.


Roming22:
Forcing merge

PR #232, opened 2024-09-16
RHTAPINST-99: Upstream MinIO Operator

otaviof:
Replacing OpenShift Operator Hub MinIO subscription with their upstream Helm Chart deployment.

PR #251, opened 2024-10-01
fix: only create tarball when necessary

Roming22:


Makefile
```diff
@@ -38,7 +38,11 @@ INSTALLER_DIR ?= ./installer
 # Tarball with the installer resources.
 INSTALLER_TARBALL ?= $(INSTALLER_DIR)/installer.tar
 # Data to include in the tarball.
-INSTALLER_TARBALL_DATA ?= charts config.yaml scripts
+INSTALLER_TARBALL_DATA ?= $(shell find $(INSTALLER_DIR) -maxdepth 1 \
+	! -path "$(INSTALLER_DIR)" \
+	! -path "$(INSTALLER_TARBALL)" \
+	! -name embed.go \
+)
```
otaviof, line 45: The tarball should always be created to avoid stable charts testing and building the project, the `tar` command only takes milliseconds to complete.

otaviof (changes requested):
We need to allow the tarball to be updated preemptively to avoid stale charts.

Makefile
Roming22: I don't understand your comment.

- After a clone, the tarball is not present -> the archive will be created at build time
- If a user modify an installer file -> the archive will be created at build time
- If a user checks out a branch with a different installer content -> the archive will be created at build time

otaviof: Also, we have `make run` (`go run`) which needs the tarball updated beforehand.

otaviof: The target `make test` also need the tarball updated, btw.

Roming22: `make test` does update the tarball.
`make run` will with the merge of #232

otaviof (approved):
Thanks for the sync up.

/lgtm

PR #252, opened 2024-10-01
fix: allow kubectl and oc in DH post-hook

Roming22:
This avoids having to download an unnecessary binary


otaviof (changes requested):
We should make either `kubectl` or `oc` as a requirement to run the `rhtap-cli` command-line, instead of trying to download something from the internet on behalf of the user.

When using `rhtap-cli` container image, it should contain all necessary requirements pre-installed.

otaviof (approved):
Latest changes looks good.

/lgtm

PR #260, opened 2024-10-04
feat: split ACS from backing-services

Roming22:


Roming22:
@otaviof CI seems to be failing on TAS install because of a Keycloak issue. Does that ring a bell?
```
(extract from rhtap-tas_rhtap-tas-test-1_rhtap-tas-test-1-realm.log)
# Checking if KeycloakRealmImport 'trusted-artifact-signer' has errors...
# KeycloakRealmImport 'trusted-artifact-signer' condition='HasErrors=True'
### [10/10] Waiting for 10 seconds before retrying...

# [ERROR] KeycloakRealmImports not available!
```

FYI, I don't see KeycloakRealmImports resources in the must-gather.

PR #262, opened 2024-10-09
fix the issue when Quay is not installed as part of rhtap

xinredhat:


xinredhat:
This fix doesn't impact the `rhtap-cli-e2e` ci in this repo. It fixes the issue that when Quay is not installed as part of RHTAP. If possible, can we merge it directly without waiting for all CI complete?

PR #266, opened 2024-10-11
fix the issue of sprayproxy-deprovision

xinredhat:
The PR https://github.com/konflux-ci/konflux-qe-definitions/pull/62 fixed issue of unregister pac server from Sprayprox. It shouldn't delete all of pac servers. This pr is to adapt that PR by passing  ocp-login-command as task parameter


Roming22 (approved):
/lgtm

Roming22:
I saw that we are failing with
```
[User error] Validation failed for pipelinerun rhtap-cli-e2e-hhq4m with error invalid input params for task sprayproxy-unregister-server: missing values for these params which have no default values: [ocp-login-command]
```
and this PR fixes that. Force merging.

PR #288, opened 2024-10-25
RHTAPINST-166: extract login logic from rhtap-install task

rhopp:


rhopp:
Pipelienes succeeded when pointing to my branch:
```
        taskRef:
          params:
            - name: url
              value: 'https://github.com/rhopp/rhtap-cli.git'
            - name: revision
              value: RHTAPINST-166
            - name: pathInRepo
              value: integration-tests/tasks/rhtap-install.yaml
```
I'll remove the temp commit and then this PR can be merged once reviewed.

rhopp:
This PR is now failing, because it's adding ENV variable which is not in main branch. 
The PR check runs with pipeline from this PR branch, task rhtap-install from `main` branch and the `install.sh` and `minio.sh` scripts again from PR branch. This is never gonna work, but I've tested it with temporary commit, that ensured the PR check runs with the task from PR branch and the test succeeded. 
I think we don't have any other option than to push even with failing PR check.

PR #299, opened 2024-11-05
Update ACS to 4.6 and enable scannerV4

Roming22:
Recommended configuration after reaching out the ACS team regarding [RHTAPBUGS-1316](https://issues.redhat.com//browse/RHTAPBUGS-1316)

### DIFF
--- a/installer/charts/rhtap-acs-scanner/scripts/test-acs-image-scan.sh
+++ b/installer/charts/rhtap-acs-scanner/scripts/test-acs-image-scan.sh
@@ -1,0 +1,1 @@
+../../../scripts/test-acs-image-scan.sh
--- a/installer/charts/rhtap-acs/templates/tests/test.yaml
+++ b/installer/charts/rhtap-acs/templates/tests/test.yaml
@@ -32,6 +32,11 @@ spec:
           value: {{ $namespace }}
         - name: RESOURCE_TYPE
           value: "deployment"
+      {{- if eq . "scanner" }}
+        # Downloading the vulnerability DB can take a long time
+        - name: RETRIES
+          value: "70"
+      {{- end }}
       command:
         - /scripts/test-rollout-status.sh
       args:
--- a/installer/charts/rhtap-acs-scanner/.helmignore
+++ b/installer/charts/rhtap-acs-scanner/.helmignore
@@ -1,0 +1,17 @@
+.bzr/
+.bzrignore
+.DS_Store
+.git/
+.gitignore
+.hg/
+.hgignore
+.idea/
+.project
+.svn/
+.vscode/
+*.bak
+*.orig
+*.swp
+*.tmp
+*.tmproj
+*~
--- a/installer/charts/rhtap-acs-scanner/Chart.yaml
+++ b/installer/charts/rhtap-acs-scanner/Chart.yaml
@@ -1,0 +1,6 @@
+---
+apiVersion: v2
+name: rhtap-acs-scanner
+description: RHTAP Advanced Cluster Security - Scanner check
+type: application
+version: 0.1.0
--- a/installer/charts/rhtap-acs-scanner/README.md
+++ b/installer/charts/rhtap-acs-scanner/README.md
@@ -1,0 +1,6 @@
+# ACS Scanner
+
+This chart does not deploy any resources, and waits on the initialization of the vulnerability database.
+
+The initialization started when `rhtap-acs` was deployed.
+Since it can take upward of 30m for the scanner to be ready, splitting this check in its own chart and running it at the end of the deployment cuts down on deployment time by allowing the install of the other products in parallel.
--- a/installer/charts/rhtap-acs-scanner/templates/_copy-scripts.tpl
+++ b/installer/charts/rhtap-acs-scanner/templates/_copy-scripts.tpl
@@ -1,0 +1,25 @@
+{{/*
+
+  POD container spec to copy scripts.
+
+*/}}
+{{- define "acs-scanner.copyScripts" -}}
+- name: copy-scripts
+  image: registry.access.redhat.com/ubi8/ubi-minimal:latest
+  workingDir: /scripts
+  command:
+    - /bin/bash
+    - -c
+    - |
+      set -x -e
+  {{- range $path, $content := .Files.Glob "scripts/*.sh" -}}
+    {{- $script := trimPrefix "scripts/" $path }}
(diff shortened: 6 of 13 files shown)


installer/charts/rhtap-acs-scanner/scripts/test-acs-image-scan.sh
```diff
@@ -0,0 +1,46 @@
...
+set -o pipefail
+
+set -x
+
+ROX_CENTRAL_ENDPOINT="$(kubectl get secrets -n rhtap rhtap-acs-integration -o jsonpath='{.data.endpoint}' | base64 -d)"
+ROX_API_TOKEN="$(kubectl get secrets -n rhtap rhtap-acs-integration -o jsonpath='{.data.token}' | base64 -d)"
+export ROX_CENTRAL_ENDPOINT
+export ROX_API_TOKEN
+
+IMAGE="${IMAGE:-quay.io/fedora/fedora:36-x86_64@sha256:d6e4c7d6d1eaa24d71c8efd7432890acdc0179502224d0aaad6bb05d15ffde19}"
```
lcarva: Oh?

lcarva: I understand it's only used for testing the installation, but can we use an image from registry.redhat.io instead?

Roming22: Unfortunately that's more complicated than just switching the image, as ACS would need to be setup to be integrated with registry.redhat.io (it's basically currently returning an error saying it can't get the image from that registry).

I've chosen as close to possible to a "Red Hat" image by using a Fedora image hosted on quay, which are 2 products strongly linked to our brand. Let me know if you don't think that's good enough.

installer/charts/rhtap-acs-scanner/scripts/test-acs-image-scan.sh
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env bash
+set -o errexit
+set -o nounset
+set -o pipefail
+
+ROX_CENTRAL_ENDPOINT="$(kubectl get secrets -n rhtap rhtap-acs-integration -o jsonpath='{.data.endpoint}' | base64 -d)"
+ROX_API_TOKEN="$(kubectl get secrets -n rhtap rhtap-acs-integration -o jsonpath='{.data.token}' | base64 -d)"
```
otaviof: You need to mount this secret on the testing POD, by doing so the `ROX_CENTRAL_ENDPOINT` and `ROX_API_TOKEN` will be available directly in the environment.

installer/charts/rhtap-acs-scanner/scripts/test-acs-image-scan.sh
```diff
@@ -0,0 +1,44 @@
...
+      --force; then
+    break
+  fi
+  echo "Waiting"
+  echo
+  sleep 60
+  echo "Retrying"
+done
+rm ./roxctl
+echo
```
otaviof: Please use a `emptyDir: {}` mount, so at the end of the container execution all data in the mount point will be gone.

installer/charts/rhtap-acs-scanner/templates/service-account.yaml
```diff
@@ -0,0 +1,46 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ .Release.Name }}
+  namespace: {{ .Release.Namespace }}
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
```
otaviof: Does it need "write" permission to Secrets?

installer/charts/rhtap-acs/templates/tests/test.yaml
```diff
@@ -32,6 +32,11 @@ spec:
           value: {{ $namespace }}
         - name: RESOURCE_TYPE
           value: "deployment"
+      {{- if eq . "scanner" }}
+        # Downloading the vulnerability DB can take a long time
```
otaviof, line 36: By default we allow a 15 minute timeout. Should we make sure the Helm Chart installation waits longer than that?

installer/charts/rhtap-acs-scanner/scripts/test-acs-image-scan.sh
otaviof, line 1: This script should live on the `installer/scritps` directory, with a symbolic link into the Helm Chart directory, following the approach taken by all other scripts. Please move it.

otaviof (changes requested):

installer/charts/rhtap-acs-scanner/scripts/test-acs-image-scan.sh
lcarva: Thanks for looking into it! I think it's fine because:

1. The contents of the fedora image are never executed.
2. The image comes from a registry that is managed by Red Hat.

installer/charts/rhtap-acs-scanner/scripts/test-acs-image-scan.sh
Roming22: Fixed

installer/charts/rhtap-acs-scanner/scripts/test-acs-image-scan.sh
Roming22: Fixed

installer/charts/rhtap-acs-scanner/templates/service-account.yaml
Roming22: In fact, we don't need the SA at all.
I now use the `default` SA in the `rhtap` namespace.

installer/charts/rhtap-acs-scanner/scripts/test-acs-image-scan.sh
Roming22: Fixed

installer/charts/rhtap-acs/templates/tests/test.yaml
Roming22: My tests shows that 15m is enough for the test to pass, we should be OK.

Roming22: Another test shows that the 15m Helm timeout does not apply to tests. I'm surprised by that result. There is a timeout flag (https://helm.sh/docs/helm/helm_test/#options), I wonder if we use it at all.

otaviof (approved):
/lgtm

PR #300, opened 2024-11-07
install.sh: set vars from files only if needed

chmeliik:
For my RHTAP development workflow, I want to disable some features and leave the related environment variables unset. Currently, I can't do that because the install script will try to read the values for those variables from files in /usr/local/rhtap-cli-install.

Change the script to read from /usr/local/rhtap-cli-install only if the variable is unset AND the corresponding feature is enabled.


chmeliik:
Looks like all the tests finally passed :tada:

Roming22 (approved):

PR #309, opened 2024-11-15
RHTAPINST-140: Go Builder Image v1.22.5

otaviof:
One image was left behind, fixing in this PR.


Roming22:
@otaviof Your PR seems to be breaking the arm64 build. I see WARNING: image platform (linux/amd64) does not match the expected platform (linux/arm64) so I believe that you put the SHA of the amd64 image instead of the manifest that has all images.

Roming22 (approved):
/lgtm

Roming22:
Forcing merge as the failure is unrelated to the change and the change (building the binary) has been successfully tested on the other tests.

PR #310, opened 2024-11-15
RHTAP-3321: Avoid race condition while patching

prietyc123:
The race condition while patching left `imagePullSecrets` with the last iteration


Roming22 (commented):
That patch seems a bit convoluted. Do you think it could be simplified a bit, eg by applying a single patch to the SA instead of multiple?

prietyc123:
> That patch seems a bit convoluted. Do you think it could be simplified a bit, eg by applying a single patch to the SA instead of multiple?

@Roming22 I have merged into a single SA patch and tested it as well. Works as expected. However I also tried applying `--type='merge'` but that also leads the race condition, so I opted to append using jq . But yes I do agree patching in multiple makes it clumsy, hence simplified it in one.

xinredhat (approved):
/lgtm

Roming22 (approved):

PR #319, opened 2024-11-21
chore: fix some SonarQube issues

Roming22:


pkg/engine/funcs_test.go
```diff
@@ -6,17 +6,21 @@ import (
 	"testing"
 )
 
+var empty_input = "empty input"
+var valid_input = "valid input"
+var invalid_input = "invalid input"
```
otaviof: Could you please move these variables into `TestToYAML` function and use camel-case? You may as well skip the word `var` by simply using `emptyInput := "empty input"` respectively.

otaviof (changes requested):

pkg/engine/funcs_test.go
Roming22: As discussed, the vars need to be global. I did rename them to camel-case.

otaviof (approved):
/lgtm

otaviof (approved):
/lgtm

otaviof (approved):
/lgtm

PR #322, opened 2024-11-25
RHTAPINST-181: MinIO Operator Namespace

otaviof:
Even with `.minIOOperator.enabled` set to `false`, the installer requires the specific namespace to be present.


jkopriva:
lets wait for CI

PR #324, opened 2024-11-26
RHTAP-2558 - Setup dynamic e2e IntegrationTestingScenario

rhopp:
This PR adds new Pipeline to be used as IntegrationTestScenario. This pipeline will allow us to specify "test plan" in form of Pict model file. This new pipeline reads this file, generates test configuration matrix based on that model and then runs full E2E scenario (from cluster creation to cluster teardown) for each config.

Current logic only supports specifying different cluster versions (other parameters from the pict model are basically ignored). This functionality will be added in future in different JIRAs/PRs.

For more info see `integration-tests/README.md`

### DIFF
--- a/integration-tests/pipelines/e2e-main-pipeline.yaml
+++ b/integration-tests/pipelines/e2e-main-pipeline.yaml
@@ -87,7 +87,7 @@ spec:
               set -x
               echo "Running tests for OCP versions:"
               echo "$CONFIGS_JSON" | jq -r '.[][] | select(.key == "OCP").value'
-              failed_pipelines=0
+              pids=()
 
               while IFS= read -r version; do
                 (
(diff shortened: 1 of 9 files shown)


integration-tests/pipelines/e2e-main-pipeline.yaml
```diff
@@ -0,0 +1,127 @@
...
+          - name: configs-json
+            description: Config json generated from pict file.
+        steps:
+          - name: pict-generate
+            image: quay.io/apodhrad/pict:latest
+            script: |
+              #!/usr/bin/env bash
+              echo "$(tasks.get-pict-file.results.pict-file)" > main.pict
+              cat main.pict
+              pict ./main.pict -o:1 -f:json > $(results.configs-json.path)
```
jsmid1, line 66: Could we make the order of combinations into a param?

rhopp: Definitely. I would just prefer to do that in followup tasks/PRs... (there is lot of room for all kinds of improvements still)

rhopp:
The tests are expected to fail (they succeeded when I had the temp commits here - which are now removed).
I'll merge this PR now.

PR #325, opened 2024-11-28
updated plr to align with release-1.3

xinredhat:


xinredhat:
@otaviof could you approve it ?

PR #341, opened 2024-12-05
RHTAPINST-186: GitHub Release Action vs. Container Image 

otaviof:
Fixing `buildah` GitHub Action error, and skipping container build in GHA given we need credentials to fetch Red Hat images.


jduimovich (approved):
This is giid. You cannot grab a secret in an action, it needs to be the workflow. 
Change looks good to me (I made the same ones)

PR #355, opened 2024-12-11
RHTAPINST-57: Show Global Values as YAML

otaviof:
By default, showing global values in the `rhtap-cli template`  as YAML, while the `--debug` flag shows key-value properties.

### Examples

-  Only showing the global values as YAML.
```bash
rhtap-cli template --show-manifests=false
```
- Rendering only the templates of a single Helm Chart.
```bash
rhtap-cli template --show-values=false charts/rhtap-subscriptions
```
- Rendering all resources of a Helm Chart.
```bash
rhtap-cli template charts/rhtap-subscriptions
```


Roming22 (approved):
/lgtm

PR #358, opened 2024-12-16
feat: add jq, oc binaries to the image

Roming22:
jq is required by the post-deploy hook for DH.
oc allows the user to connect to the cluster from within the image.


otaviof (approved):
/lgtm

PR #367, opened 2025-01-03
Do not stream logs. Print URL to nested PLR instead.

rhopp:
This PR removes the log streaming from nested PLRs to clean up the output. Instead of that it will print link to the nested pipelineruns.

### DIFF
--- a/integration-tests/pipelines/e2e-main-pipeline.yaml
+++ b/integration-tests/pipelines/e2e-main-pipeline.yaml
@@ -100,7 +100,7 @@ spec:
                     fieldPath: metadata.namespace
             script: |
               #!/usr/bin/env bash
-              set -euxo pipefail
+              set -euo pipefail
 
               GIT_REPO="$(jq -r '.git.repo // empty' <<< $JOB_SPEC)"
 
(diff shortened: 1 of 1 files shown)


integration-tests/pipelines/e2e-main-pipeline.yaml
```diff
@@ -141,25 +144,50 @@ spec:
...
-                return_code=$?
-                if [ $return_code -ne 0 ]; then
-                  echo "One or more pipelineruns failed. Exiting with non-zero code."
-                  oc get pipelineruns
-                  exit 1
+
+              # Wait for all PLRs to finish 
+              timeout --foreground 120m /bin/bash -c '
+                  while true; do
+                    IS_SOMETHING_RUNNING=false
```
prietyc123:
```suggestion
                    IS_PIPELINES_RUNNING=false
```

integration-tests/pipelines/e2e-main-pipeline.yaml
```diff
@@ -141,25 +144,50 @@ spec:
...
+                    if $IS_SOMETHING_RUNNING ; then
+                      echo "Nested pipelines are still running. Waiting 1 minute"
+                      sleep 60
+                    else
+                      echo "All nested pipelines finished"
+                      break
+                  done
+              '
+              ## Explore and report status of all failed pipelineruns. Fail if anything failed
+              SOMETHING_FAILED=false
```
prietyc123:
```suggestion
              IS_PIPELINE_FAILED=false
```

integration-tests/pipelines/e2e-main-pipeline.yaml
```diff
@@ -141,25 +144,52 @@ spec:
...
+
+              # Wait for all PLRs to finish
+              timeout --foreground 120m /bin/bash -c '
+                  #deserialize plrs array
+                  read -r -a PIPELINERUNS_ARRAY <<< "$PIPELINERUNS_ARRAY_SERIALIZED"
+                  echo "${PIPELINERUNS_ARRAY[*]}"
+                  while true; do
+                    ARE_PIPELINES_RUNNING=false
+                    for PIPELINE_RUN in "${PIPELINERUNS_ARRAY[@]}"; do
+                      if [[ $(oc get pipelinerun/$PIPELINE_RUN -n ${KONFLUX_NAMESPACE} -o jsonpath="{.status.conditions[?(@.type==\"Succeeded\")].status}") == "Unknown" ]]; then
```
lcarva, line 167: Have you tried `oc/kubectl wait`? Not sure if it is possible, but might be worth checking out.

(This is definitely not a blocker comment.)

lcarva (approved):

integration-tests/pipelines/e2e-main-pipeline.yaml
rhopp: I was looking into that as well, but `wait` waits for a condition to have certain status. By default it's waiting for the status to be `true` (you can specify different value). But in this case, the PLR, once finished can be either `True` or `False` (https://tekton.dev/docs/pipelines/pipelineruns/#monitoring-execution-status). 
Instead of that I've created this cycle to actively wait for the status to be *not* Unknown which I don't think is possible with `oc wait`.

lcarva: Bummer. Thanks for checking!

PR #368, opened 2025-01-03
fix: RHTAP-3416 'template' command uses the wrong namespace

Roming22:
- The chart release namespace is retrieved from the config.yaml.
- Helm connects to the server to provide an accurate rendering of the templates.

rh-pre-commit.version: 2.3.2

### DIFF
--- a/installer/charts/rhtap-dh/values.yaml
+++ b/installer/charts/rhtap-dh/values.yaml
@@ -1,4 +1,7 @@
 ---
+debug:
+  ci: false
+
 developerHub:
   instanceName: developer-hub
   catalogURL: __OVERWRITE_ME__
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -55,6 +55,17 @@ func (c *Config) GetBaseDir() string {
 	return filepath.Dir(c.configPath)
 }
 
+// GetDependency returns a dependency chart configuration.
+func (c *Config) GetDependency(logger *slog.Logger, chart string) (*Dependency, error) {
+	logger.Debug("Getting dependency")
+	for _, dep := range c.Installer.Dependencies {
+		if dep.Chart == chart {
+			return &dep, nil
+		}
+	}
+	return nil, fmt.Errorf("chart %s not found", chart)
+}
+
 // GetEnabledDependencies returns a list of enabled dependencies.
 func (c *Config) GetEnabledDependencies(logger *slog.Logger) []Dependency {
 	enabled := []Dependency{}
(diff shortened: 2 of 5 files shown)


pkg/deployer/helm.go
```diff
@@ -68,7 +68,7 @@ func (h *Helm) helmInstall(vals chartutil.Values) (*release.Release, error) {
 	c.DryRun = h.flags.DryRun
 	c.ClientOnly = h.flags.DryRun
 	if h.flags.DryRun {
-		c.DryRunOption = "client"
+		c.DryRunOption = "server"
```
otaviof, line 71: Why are we changing this from `client` to `server`?

installer/charts/rhtap-dh/values.yaml
```diff
@@ -1,4 +1,7 @@
 ---
+debug:
+  ci: false
+
```
otaviof, line 4: Is the `rhtap-dh` Chart using `.debug.ci` in the templates? If I remember correctly, the other charts don't have this attribute defined.

pkg/config/config.go
```diff
@@ -55,6 +55,17 @@ func (c *Config) GetBaseDir() string {
 	return filepath.Dir(c.configPath)
 }
 
+// GetChart returns a dependency chart configuration.
+func (c *Config) GetChart(logger *slog.Logger, chart string) (*Dependency, error) {
```
otaviof: Given this method returns a `&Dependency{}`, let's name it `GetDependency` instead?

otaviof (changes requested):
Let me understand the reasons behind the dry-run changes, please.

installer/charts/rhtap-dh/values.yaml
Roming22: That's the only template that uses that flag.
I've added it because it allows to call `helm template` on the chart without any error.

pkg/deployer/helm.go
Roming22: Because a 'client' rendering is mostly useless for troubleshooting. The output will not be representative of what will be deployed on the cluster.

Switching to 'server' allows Helm to contact the server and produce the manifests as they will be applied at deployment time.

pkg/config/config.go
Roming22: With your latest changes, I think I can rollback that change.

Roming22: It is in fact still needed so I've renamed it.

Roming22:
@otaviof I've fixed your comment, and answered your questions. I've also added an extra commit to fix an issue created by a recent commit.

PTAL.

pkg/deployer/helm.go
otaviof, line 71: By using `client` means Helm dry-run mode validates configurations locally, while `server` dry-run mode sends the request to the API server for full validation without persisting changes. We can improve this later, let's highlight that 👍

otaviof (approved):
/lgtm

With a single remark about the dry-run mode 😊

PR #371, opened 2025-01-07
added support for nexus and artifacory

xinredhat:


integration-tests/scripts/install.sh
```diff
@@ -54,7 +56,17 @@ export BOMBASTIC_API_URL
...
+export ARTIFACTORY_URL
+export ARTIFACTORY_TOKEN
+export ARTIFACTORY_DOCKERCONFIGJSON
+export ARTIFACTORY_USERNAME
+export ARTIFACTORY_PASSWORD
+## variables for Nexus
+export NEXUS_URL
+export NEXUS_DOCKERCONFIGJSON
+export NEXUS_USERNAME
+export NEXUS_PASSWORD
```
prietyc123: Too many env variable. Can we just have all these into a file and export the file content in this script

integration-tests/scripts/install.sh
```diff
@@ -153,24 +120,57 @@ acs_integration() {
...
+  else
     acs_central_url=https://$(kubectl -n rhtap-acs get route central -o  'jsonpath={.spec.host}')
     acs_central_password=$(kubectl -n rhtap-acs get secret central-htpasswd -o go-template='{{index .data "password" | base64decode}}')
+  fi
+
+  # if registry_config is "quay", then perform the quay integration
+  if [[ "${registry_config}" == "quay" ]]; then
+    echo "[INFO] Configure Quay integration into ACS"
+
     quay_host=$(kubectl -n rhtap-quay get route rhtap-quay-quay -o  'jsonpath={.spec.host}')
```
prietyc123, line 139: In case of hosted quay this value won't exists, no?

integration-tests/scripts/install.sh
```diff
@@ -201,34 +201,62 @@ tpa_integration() {
...
   jenkins_integration
-  gitlab_integration
-  quayio_integration
+  tpa_integration
   acs_integration
+  github_integration
+  gitlab_integration
   bitbucket_integration
-  tpa_integration
+  quayio_integration
```
prietyc123, line 252: I don't see this function in this file, neither gitlab_integration, lots of them are missing from my eyes. Am I missing something. Damn 🙄

prietyc123: Is the assumption is we will only run for the rhtap installed one. No hosted quay?

integration-tests/scripts/install.sh
```diff
@@ -201,34 +201,62 @@ tpa_integration() {
   fi
 }
 
+artifactory_integration() {
+  if [[ "${registry_config}" == "artifactory" ]]; then
+    echo "[INFO] Configure Artifactory integration into RHTAP"
+    # disable Quay installation
+    yq e '.rhtapCLI.features.redHatQuay.enabled = false' -i "${config_file}"
```
prietyc123, line 219: what happens when quay is enabled then artifactory and nexus secret is also there?

integration-tests/scripts/install.sh
xinredhat: oh, In case of hosted quay(quay.io), when ACS is newly created we need to configure hosted quay integration with ACS

integration-tests/scripts/install.sh
xinredhat: if you just run the install.sh to install rhtap, this case never happens. If we need the case you mention, then we need to consider if we need to support rhtap with mulitiple Container Registries, and SCM, etc.

PR #372, opened 2025-01-10
Update owners files

rhopp:
* Replace John with @lcarva 
* Add @rhopp and @jkopriva to reviewers in root dir
* Add QE folks as approvers in /integration-tests/ dir


lcarva (approved):
Thanks!

PR #384, opened 2025-01-15
Fix: Fail pipelinerun if any of nested pipelines fails.

rhopp:
There is some mistake in the logic - Script is not waiting correctly for the "nested" pipelineruns to finish.


rhopp:
/retest
;

jkopriva (approved):

PR #385, opened 2025-01-15
feat: split DH instance config and gitops target namespaces

Roming22:
This change helps with testing the scenario when gitops is deployed separately from RHTAP.

- Creation of a new rhtap-app-namespaces dependency responsible for the setup of the namespaces where applications are built and deployed.
- The new rhtap-app-namespaces dependency is controlled by Pipelines, as they are used to run the pipelines and applications. It is not controlled by GitOps, as in the future the GitOps instance could live on a different cluster.

rh-pre-commit.version: 2.3.2


installer/charts/rhtap-gitops/values.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 argoCD:
   # Toggles the ArgoCD service.
-  enabled: false
+  enabled: true
```
otaviof: By default charts should be disabled, and explicitly enabled in the `config.yaml` or via `values.yaml.tpl` output.

otaviof (changes requested):
Minor request for change, other than that looks good to me.

installer/charts/rhtap-gitops/values.yaml
Roming22: I had a reason to set it to `true`, but it's outside the scope of this PR. So I'll revert the change, and we can talk later on about what led me to this change.

otaviof (approved):
/lgtm

PR #387, opened 2025-01-16
fix: RHTAP-4012 Container improvements

Roming22:
- Add 'vi' to enable users to edit files
- Change files ownership so users can edit files
- Remove unnecessary build files from the container

rh-pre-commit.version: 2.3.2


Roming22:
I needed to change the UID because of a new build error: `useradd warning: rhtap-cli's uid 1000 is greater than SYS_UID_MAX 999`

otaviof (approved):
/lgtm

PR #396, opened 2025-01-21
fix: bump TSSC template to 1.3.0

Roming22:
Use the latest stable version of the templates by default.

rh-pre-commit.version: 2.3.2


installer/config.yaml
```diff
@@ -17,7 +17,7 @@ rhtapCLI:
       enabled: &rhdhEnabled true
       namespace: *installerNamespace
       properties:
-        catalogURL: https://github.com/redhat-appstudio/tssc-sample-templates/blob/release/all.yaml
+        catalogURL: https://github.com/redhat-appstudio/tssc-sample-templates/blob/v1.3.0/all.yaml
```
lcarva, line 20: I wonder if for the main branch of rhtap-cli, this should use the main branch of tssc-sample-templates?

Presumably, this will have to change regardless for 1.4?

Roming22: I don't want to use the main branch by default because in the past we had unexpected failure when the main branch was in a bad state.

jkopriva (approved):

Roming22:
Can be merged if test is successful on a least one OCP instance.

PR #415, opened 2025-01-30
fix: template subcommand

Roming22:
The template subcommand was behaving differently if the chart was already deployed.

This is now corrected to behave the same as if the chart was to be installed, which gives an accurate representation of the manifests that Helm will apply.


otaviof (commented):
/lgtm

PR #416, opened 2025-01-30
chore: ignore symlinks to .vscode

Roming22:
Developers might use symlinks to the directory, so the change ignore both the dir and the file.


otaviof (commented):
/lgtm

PR #433, opened 2025-02-04
feat: Add option to bypass subscriptions creation

Roming22:
Clusters might already have subscriptions installed, which would trigger an error in Helm as it did not control the resources.

A new flag is added to disable the subscription installation per product, with the following limitations:
- AMQ Streams is controlled by TPA
- CrunchyData is controlled by RHDH and TPA

Those limitations means that a user may need to install some subscription manually (eg if they already have AMQ Streams, TPA must be installed manually and the `managedSubscription` flag must be set to `false`).

cf [RHTAP-4063](https://issues.redhat.com//browse/RHTAP-4063)

rh-pre-commit.version: 2.3.2

### DIFF
--- a/installer/charts/rhtap-subscriptions/templates/_subscriptions.tpl
+++ b/installer/charts/rhtap-subscriptions/templates/_subscriptions.tpl
@@ -10,3 +10,15 @@
   {{- end -}}
   {{- $enabled | toYaml -}}
 {{- end -}}
+{{- /*
+  Select the managed subscriptions.
+*/ -}}
+{{- define "subscriptions.managed" -}}
+  {{- $managed := dict -}}
+  {{- range $k, $v := .Values.subscriptions -}}
+    {{- if $v.managed -}}
+      {{- $managed = merge $managed (dict $k $v) -}}
+    {{- end -}}
+  {{- end -}}
+  {{- $managed | toYaml -}}
+{{- end -}}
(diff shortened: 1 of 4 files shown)


installer/charts/rhtap-subscriptions/values.yaml
```diff
@@ -22,7 +22,7 @@ subscriptions:
     enabled: false
     description: OpenShift GitOps Operator
     apiResource: gitopsservices.pipelines.openshift.io
-    namespace: openshift-operators
+    namespace: openshift-gitops-operator
```
otaviof: Interesting. Back in the days, I had issues on moving subscription operators outside of `openshift-operators` namespace.

otaviof (approved):
Nice one! I liked the approach exploring the feature properties dictionary 👍

/lgtm

installer/charts/rhtap-subscriptions/values.yaml
Roming22: I've changed it because it's the default namespace used if you install the subscription from the console. I thought it would be a good idea to be as vanilla as possible.

Roming22: I didn't test it properly. You are right, it doesn't work as OCP does not allow the creation of `openshift-*` namespaces.

I'll revert this change.

Roming22: Reverted.

PR #473, opened 2025-02-18
fix case where the PLR was not yet created

rhopp:
Sometimes the main pipeline ends & succeeds even though the plrs it created are still running. I suppose it's because when this IF condition is evaluated, the PLR is not yet existing. This PR should take care of that.


jkopriva (approved):

Roming22 (approved):
/lgtm

PR #479, opened 2025-02-20
Path service-account.yaml.patch.yaml does not exist error

lingyzhuang:
If none of the three secrets: `artifactory-auth nexus-auth quay-auth` exists, the script would fail with error: `error: the path service-account.yaml.patch.yaml does not exist`.


Roming22 (approved):
/lgtm

Roming22:
Force merge, as the CI is unstable and tests are successful on OCP 4.16.

PR #485, opened 2025-02-26
RHTAP-4048: `rhtap-cli config`

otaviof:
This change allows all cluster administrators to rely on a single configuration.

Introduces the `rhtap-cli config` subcommand, responsible for managing a unique installer's configuration in the cluster.

The configuration is then used from the cluster for all other subcommands, thus it becomes the first step on deploying RHTAP.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -4,21 +4,11 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
-	"path/filepath"
 
 	"github.com/redhat-appstudio/rhtap-cli/pkg/chartfs"
-	"github.com/spf13/pflag"
+
 	"gopkg.in/yaml.v3"
 )
-
-// ErrInvalidConfig indicates the configuration content is invalid.
-var ErrInvalidConfig = errors.New("invalid configuration")
-
-// ErrEmptyConfig indicates the configuration file is empty.
-var ErrEmptyConfig = errors.New("empty configuration")
-
-// ErrUnmarshalConfig indicates the configuration file structure is invalid.
-var ErrUnmarshalConfig = errors.New("failed to unmarshal configuration")
 
 // Spec contains all configuration sections.
 type Spec struct {
@@ -34,26 +24,23 @@ type Spec struct {
 
 // Config root configuration structure.
 type Config struct {
-	cfs        *chartfs.ChartFS // embedded filesystem
-	configPath string           // configuration file path
+	cfs     *chartfs.ChartFS // embedded filesystem
+	payload []byte           // original configuration payload
 
 	Installer Spec `yaml:"rhtapCLI"` // root configuration for the installer
 }
 
-// PersistentFlags defines the persistent flags for the CLI.
-func (c *Config) PersistentFlags(f *pflag.FlagSet) {
-	f.StringVar(
-		&c.configPath,
-		"config",
-		c.configPath,
-		"Path to the installer configuration file",
-	)
-}
+var (
+	// ErrInvalidConfig indicates the configuration content is invalid.
+	ErrInvalidConfig = errors.New("invalid configuration")
+	// ErrEmptyConfig indicates the configuration file is empty.
+	ErrEmptyConfig = errors.New("empty configuration")
+	// ErrUnmarshalConfig indicates the configuration file structure is invalid.
+	ErrUnmarshalConfig = errors.New("failed to unmarshal configuration")
+)
 
-// GetBaseDir returns the base directory of the configuration file.
-func (c *Config) GetBaseDir() string {
-	return filepath.Dir(c.configPath)
-}
+// DefaultRelativeConfigPath default relative path to YAML configuration file.
+var DefaultRelativeConfigPath = fmt.Sprintf("installer/%s", Filename)
 
 // GetDependency returns a dependency chart configuration.
 func (c *Config) GetDependency(logger *slog.Logger, chart string) (*Dependency, error) {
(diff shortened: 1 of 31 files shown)


pkg/subcmd/config.go
```diff
@@ -0,0 +1,240 @@
...
+// log returns a decorated logger.
+func (c *Config) log() *slog.Logger {
+	return c.logger.With("config-path", c.configPath)
+}
+
+// PersistentFlags injects the sub-command flags.
+func (c *Config) PersistentFlags(p *pflag.FlagSet) {
+	p.BoolVarP(
+		&c.create,
+		"apply",
```
Roming22: Should most likely be renamed to "create".
We could have "apply", add a "--force" flag and remove "update", so it behaves like the integrations.

pkg/integrations/acs.go
```diff
@@ -66,29 +65,40 @@ func (a *ACSIntegration) Validate() error {
...
+	feature, err := cfg.GetFeature(config.RedHatDeveloperHub)
+	if err != nil {
+		return err
+	}
 	return k8s.EnsureOpenShiftProject(
 		ctx,
 		a.log(),
 		a.kube,
-		a.cfg.Installer.Namespace,
+		feature.GetNamespace(),
```
Roming22: You were a bit hasty in the resolving the conflicts. Integrations secrets goes in the installer namespace, not the DH namespace ever since DH/gitops got their own namespace.

pkg/config/config.go
```diff
@@ -34,26 +25,22 @@ type Spec struct {
...
-		"Path to the installer configuration file",
-	)
-}
+var (
+	// ErrInvalidConfig indicates the configuration content is invalid.
+	ErrInvalidConfig = errors.New("invalid configuration")
+	// ErrEmptyConfig indicates the configuration file is empty.
+	ErrEmptyConfig = errors.New("empty configuration")
+	// ErrUnmarshalConfig indicates the configuration file structure is invalid.
+	ErrUnmarshalConfig = errors.New("failed to unmarshal configuration")
```
Roming22, line 39: `marshall` and `unmarshall` (typo appears multiple times in the code, please fix all instances)

pkg/config/manager.go
```diff
@@ -0,0 +1,183 @@
...
+
+	// When no ConfigMaps matching criteria is found in the cluster.
+	if len(configMapList.Items) == 0 {
+		return nil, fmt.Errorf(
+			"%w: using label selector %q",
+			ErrConfigMapNotFound,
+			m.selectorLabel(),
+		)
+	}
+	// Also, important to error out when multiple ConfigMaps are present in the
```
Roming22, line 68: In the cluster or in the namespace? I could imagine multiple RHTAP instances in the same cluster.

pkg/subcmd/config.go
```diff
@@ -0,0 +1,240 @@
...
+		payload, err := cfg.String()
+		if err != nil {
+			return err
+		}
+		fmt.Print(payload)
+	}
+	return nil
+}
+
+// NewConfig instantiates the a new "config" subcommand.
```
Roming22: `the a`?

pkg/subcmd/config.go
```diff
@@ -0,0 +1,240 @@
...
+	if c.get && c.delete {
+		return fmt.Errorf("cannot get and delete at the same time")
+	}
+	if !c.create && !c.update && !c.get && !c.delete {
+		return fmt.Errorf("either apply, update, get or delete must be set")
+	}
+	return nil
+}
+
+// Complete inspect the context to determin the path of the configuration file, or
```
Roming22: `determine`

pkg/subcmd/config.go
```diff
@@ -0,0 +1,240 @@
...
+// uses the embedded payload, makes sure the args are adequate.
+func (c *Config) Complete(args []string) error {
+	// It should return error more than a single argument is informed.
```
Roming22: `an error if`

Roming22 (changes requested):
My test with ACS/Quay integrations is failing for 2 reasons:

- integrations secrets are not created in the right namespace anymore (bad conflict resolution)
- The configmap does not seem to be taken into account at deploy time (feature is disabled but is installed)

```
> rhtap-cli deploy

...
############################################################
# [6/15] Deploying 'charts/rhtap-acs' in 'rhtap-acs'.
############################################################
Error: install failed: unable to build kubernetes objects from release manifest: resource mapping not found for name: "stackrox-central-services" namespace: "" from "": no matches for kind "Central" in version "platform.stackrox.io/v1alpha1"
ensure CRDs are installed first
bash-5.1$ oc get configmap -n rhtap rhtap-cli-config -o yaml | grep -A2 redHatAdvancedClusterSecurity
            redHatAdvancedClusterSecurity:
                enabled: false
                namespace: rhtap-acs
```

pkg/config/manager.go
otaviof: For this story only a single RHTAP instance is supported in the cluster. So, in the cluster.

pkg/config/config.go
otaviof: Sure! I realise am spelling "marshal" wrong this whole time 😳

pkg/subcmd/config.go
otaviof: Sure, please consider latest changes.

pkg/config/manager.go
Roming22: Can you add the installer namespace to the List?

1. I'm uncomfortable with the config being scattered across the cluster.
2. It won't be a problem when/if we add multi RHTAP instances support.

otaviof: The config is created, by default, in the installer namespace. You need to change the `config.yaml` to have another namespace in place, in other words, uses the same namespace than the installer.

Roming22: I now understand that since the installer namespace is in the config itself, the List needs to happen across the cluster.

PR #502, opened 2025-03-03
[DO NOT MERGE] RHTAP-4516 RHDH pre-release testing

jkopriva:


Roming22:
@jkopriva I'd be interested in having that as a script in `hack/` so devs can easily deploy and test unstable versions too.

PR #505, opened 2025-03-05
Wait for starting pipelines in separate loop

rhopp:
Before this fix we sometimes get the situation, where pipelines were started and one of them failed/ended quickly and it was pruned from the cluster befor the second one finished. This got us to a state, where the loop waiting for the PLRs to finish was stuck even though the pipelines were finished.
This PR splits the condition to two wait loops - one for the pipelines to start, the second one for pipelines to finish.


Roming22 (approved):
/lgtm

PR #511, opened 2025-03-07
Adding expiration period to images pushed to quay

dperaza4dustbit:
This should help with our quay repository tag building up. I had to manually delete or set expiration on all the tags.

### DIFF
--- a/.tekton/rhtap-cli-push.yaml
+++ b/.tekton/rhtap-cli-push.yaml
@@ -27,6 +27,8 @@ spec:
     value: quay.io/rhtap/rhtap-cli:{{revision}}
   - name: dockerfile
     value: Dockerfile
+  - name: image-expires-after
+    value: 5d
   pipelineSpec:
     finally:
     - name: show-sbom


.tekton/rhtap-cli-push.yaml
```diff
@@ -27,6 +27,8 @@ spec:
     value: quay.io/rhtap/rhtap-cli:{{revision}}
   - name: dockerfile
     value: Dockerfile
+  - name: image-expires-after
+    value: 5d
```
lcarva, line 31: To be clear, these images are not releasable. That's probably ok because this is the main branch, but let's make sure the process of creating the release branches doesn't unintentionally pick this up.

I think 5 days is overly aggressive, but if that works for you, go for it.

dperaza4dustbit: going with 5d as it was what is set in PR pull PipelineRun https://github.com/redhat-appstudio/rhtap-cli/blob/main/.tekton/rhtap-cli-pull-request.yaml#L30. But will be OK with a higher number, we just want to avoid a buildup.

dperaza4dustbit: What is the process for creating a release branch, I can add steps there to make sure this is not move there

Roming22 (approved):
/lgtm

Roming22:
The release process is controlled by https://github.com/redhat-appstudio/rhtap-cli/blob/main/.github/workflows/release.yaml. If you need more info Otavio is the SME.

PR #512, opened 2025-03-10
fix: extract REGISTRY_ENDPOINT from .dockerconfigjson

xinredhat:
rh-pre-commit.version: 2.3.2

For both Nexus and Artifactory, they have two urls, one is console url, which is used to login registry UI to manage registry. Another one is registry url, which is used for pushing/pull images to registry server with podman or docker cli.  The value of `.data.url` is image registry console url, not registry url. We need to extract registry url from .dockerconfigjson

### DIFF
--- a/installer/scripts/acs-integration-helper.sh
+++ b/installer/scripts/acs-integration-helper.sh
@@ -90,13 +90,17 @@ get_variables() {
     if [ -z "${ROX_API_TOKEN:-}" ]; then
         ROX_API_TOKEN="$(oc get secrets -n "$NAMESPACE" "$ACS_SECRET" -o json | jq -r '.data.token | @base64d')"
     fi
-
+    # For Nexus and Artifactory, they have two urls, one for UI and one for image registry. When RHTAP integrate with them, 
+    # it needs to pass UI url as `--url` parameter to the integration command. If passing image registry url as `--url` parameter,
+    # images pushed to Nexus and Artifactory will not apear in the Image Registry page.
     if [ -z "${REGISTRY_ENDPOINT:-}" ]; then
-        REGISTRY_ENDPOINT="$(oc get secrets -n "$NAMESPACE" "$REGISTRY_SECRET" -o json | jq -r '.data.url | @base64d')"
+        REGISTRY_ENDPOINT="$(
+            oc get secrets -n "$NAMESPACE" "$REGISTRY_SECRET" -o json \
+            | jq -r '.data.".dockerconfigjson" | @base64d' \
+            | jq -r '.auths | to_entries[0].key'
+        )"
     fi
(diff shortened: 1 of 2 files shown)


installer/scripts/acs-integration-helper.sh
```diff
@@ -92,11 +92,13 @@ get_variables() {
     fi
 
     if [ -z "${REGISTRY_ENDPOINT:-}" ]; then
-        REGISTRY_ENDPOINT="$(oc get secrets -n "$NAMESPACE" "$REGISTRY_SECRET" -o json | jq -r '.data.url | @base64d')"
-    fi
-    if [ -z "${REGISTRY_HOST:-}" ]; then
-        REGISTRY_HOST="${REGISTRY_ENDPOINT//https:\/\//}"
+        REGISTRY_ENDPOINT="$(
```
Roming22, line 97: That should not be needed. Is the `url` missing when using some integrations? If that is the case then the integrations should be fixed.

Roming22 (changes requested):

installer/scripts/acs-integration-helper.sh
xinredhat: No, the `url` is not missing, the `url` is console url when using integration with Nexus or Artifactory. I don't think It's a bug. I verified that  if the url is the registry endpoint when integrating with Nexus or Artifactory, images cannot show up in Image Registry page.

xinredhat: I am confused why `REGISTRY_ENDPOINT ` should not be needed. It needs to be passed into the data body as the value of endpoint.
```
{
    "autogenerated": false,
    "categories": ["REGISTRY"],
    "clusterId": "",
    "id": "",
    "name": "'"${REGISTRY_ENDPOINT}"'",
    "'"${INTEGRATION_TYPE}"'": {
        "endpoint": "'"${REGISTRY_ENDPOINT}"'",
        "insecure": '"$INSECURE"',
        '"$REGISTRY_CREDENTIALS"'
    },
    "skipTestIntegration":false,
    "type":"'"${REGISTRY_TYPE}"'"
}
```

Roming22: REGISTRY_ENDPOINT is needed.

My understanding is that REGISTRY_ENDPOINT should match the `url` of the integration secret, but you are saying `I verified that if the url is the registry endpoint when integrating with Nexus or Artifactory, images cannot show up in Image Registry page.`, so I'm obviously wrong.

So now, I'm trying to understand how the `url` and `.auth` values differ. Can you show me an example for each of the image registry?

xinredhat: Added comments to explain it

Roming22 (approved):
/lgtm

PR #524, opened 2025-03-12
fix: cascade config changes to deploy.sh

Roming22:
rh-pre-commit.version: 2.3.2


Roming22:
Force merging because untested by the CI

PR #549, opened 2025-03-18
feat: update TektonChain configuration

Roming22:
Use the operator configuration to generate the signing-secret. Tests have shown that the operator does not overwrites an existing secret, so no additional safeguards are needed to enable the feature.

Remove the option to configure the secret name, as the name is static for TektonChain.

rh-pre-commit.version: 2.3.2


dperaza4dustbit (approved):
/lgtm

PR #554, opened 2025-03-20
Bind new ClusterRole to service account rhdh-kubernetes-plugin

lingyzhuang:
Add a ClusterRole rhdh-kubernetes-plugin and bind it to service account rhdh-kubernetes-plugin.
The sources of the role is copied from: https://backstage.io/docs/features/kubernetes/configuration#role-based-access-control
And added:
1. apiGroups: `argoproj.io`, `route.openshift.io` and `tekton.dev`, which is from https://github.com/redhat-appstudio/rhtap-cli/blob/1b390bf3d25c628d4b773c4c33b9ffc2e68e072d/installer/charts/rhtap-dh/templates/plugins-content.yaml#L120).
2. resource: `pods/log`, which is needed to view pods log in catalog.

Jira: https://issues.redhat.com/browse/RHTAP-4550

### DIFF
--- a/installer/charts/rhtap-infrastructure/templates/developer-hub/serviceaccount.yaml
+++ b/installer/charts/rhtap-infrastructure/templates/developer-hub/serviceaccount.yaml
@@ -33,13 +33,83 @@ secrets:
 
 ---
 apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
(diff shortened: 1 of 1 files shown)


Roming22 (commented):
/lgtm

Roming22 (approved):

PR #586, opened 2025-03-31
fix: Simplify TektonConfig management

Roming22:
The Pipelines operator can be configured to not deploy a default TektonConfig. We leverage that feature to remove the workaround we used to take control over the resource.


installer/charts/rhtap-pipelines/values.yaml
```diff
@@ -1,2 +1,3 @@
 ---
-pipelines: {}
\ No newline at end of file
+pipelines:
+  namespace: __OVERWRITE_ME__
```
dperaza4dustbit, line 3: Not understanding this value, why not setting to the default that was hardcoded in the template "openshift-pipelines"

dperaza4dustbit (commented):
Just one question, looks good otherwise

installer/charts/rhtap-pipelines/values.yaml
Roming22: This is our convention for saying "this value comes from values.yaml.tpl and/or config.yaml".
We could choose to hardcode the value in the future if there's no benefit to having it as a variable.

dperaza4dustbit: Understand now

dperaza4dustbit (approved):
/lgtm

PR #593, opened 2025-04-02
fix: 'template' subcommand fails when '--show-manifests=false' is used

Roming22:


pkg/config/config.go
```diff
@@ -50,7 +50,7 @@ func (c *Config) GetDependency(logger *slog.Logger, chart string) (*Dependency,
 			return &dep, nil
 		}
 	}
-	return nil, fmt.Errorf("chart %s not found", chart)
+	return nil, fmt.Errorf("chart '%s' not found", chart)
```
lcarva, line 53: fyi `%q` will quote strings for you.

pkg/subcmd/template.go
```diff
@@ -81,9 +81,6 @@ func (t *Template) Complete(args []string) error {
 	// to false it will return a validation error.
 	t.flags.DryRun = true
 
-	if !t.showManifests {
```
lcarva, line 84: Can you explain what/how this is fixing? :pray:

Roming22: If `--show-manifest` is set to `false`, we exit from the method before setting `t.dep.Chart`. This triggers an error because the chart is not set, but we try to retrieve it at a later point in time.

A better solution could be found, but this fix is an easy workaround, and I'm not sure of the value of spending more time designing a better solution.

lcarva (approved):
/lgtm

PR #597, opened 2025-04-03
Remove temporary serviceaccounts and clusterrolebindings

lingyzhuang:
1. Add `need-delete` labels to service accounts and clusterrolebindings
2. When deploy failed or finished, delete serviceaccounts and clusterrolebindings with label `need-delete`
3. Retry on delete operations, retry 5 times with 10 seconds waiting.
Jira: [RHTAP-4552](https://issues.redhat.com/browse/RHTAP-4552)

### DIFF
--- a/installer/charts/_common/_helpers.tpl
+++ b/installer/charts/_common/_helpers.tpl
@@ -43,6 +43,13 @@ app.kubernetes.io/managed-by: {{ .Release.Service }}
 {{- end }}
 
 {{/*
+Need deleted labels
+*/}}
+{{- define "common.postDeployDeleteLabels" -}}
+rhtap-cli.redhat-appstudio.github.com/post-deploy: delete
+{{- end }}
+
+{{/*
 Selector labels
 */}}
 {{- define "common.selectorLabels" -}}
--- a/pkg/k8s/delete_resources.go
+++ b/pkg/k8s/delete_resources.go
@@ -1,0 +1,164 @@
+package k8s
+
+import (
+	"context"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// labelSelector is the label set for resources to be deleted.
+const labelSelector = "rhtap-cli.redhat-appstudio.github.com/post-deploy=delete"
+
+// DeleteClusterRoleBindings deletes Kubernetes ClusterRoleBindings by label.
+func DeleteClusterRoleBindings(
+	ctx context.Context,
+	kube *Kube,
+	namespace string,
+) error {
+	rbacClient, err := kube.RBACV1ClientSet(namespace)
+	if err != nil {
+		return err
+	}
+	return rbacClient.ClusterRoleBindings().
+		DeleteCollection(ctx, metav1.DeleteOptions{},
+			metav1.ListOptions{LabelSelector: labelSelector})
+}
+
+// DeleteClusterRoles deletes Kubernetes ClusterRoles by label.
+func DeleteClusterRoles(
+	ctx context.Context,
+	kube *Kube,
+	namespace string,
+) error {
+	rbacClient, err := kube.RBACV1ClientSet(namespace)
+	if err != nil {
+		return err
+	}
+	return rbacClient.ClusterRoles().
+		DeleteCollection(ctx, metav1.DeleteOptions{},
+			metav1.ListOptions{LabelSelector: labelSelector})
+}
+
+// DeleteRoleBindings deletes Kuberbetes RoleBindings by label.
+func DeleteRoleBindings(
+	ctx context.Context,
+	kube *Kube,
+	namespace string,
+) error {
+	rbacClient, err := kube.RBACV1ClientSet(namespace)
+	if err != nil {
+		return err
+	}
+	RoleBindingsList, err := rbacClient.RoleBindings("").
+		List(ctx, metav1.ListOptions{LabelSelector: labelSelector})
+	for _, rb := range RoleBindingsList.Items {
+		err := rbacClient.RoleBindings(rb.Namespace).
+			Delete(ctx, rb.Name, metav1.DeleteOptions{})
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// DeleteRoles deletes Kubernetes Roles by label.
+func DeleteRoles(
+	ctx context.Context,
+	kube *Kube,
+	namespace string,
+) error {
+	rbacClient, err := kube.RBACV1ClientSet(namespace)
+	if err != nil {
+		return err
+	}
+	RolesList, err := rbacClient.Roles("").
+		List(ctx, metav1.ListOptions{LabelSelector: labelSelector})
+	for _, role := range RolesList.Items {
+		err := rbacClient.Roles(role.Namespace).
+			Delete(ctx, role.Name, metav1.DeleteOptions{})
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// DeleteServiceAccounts deletes Kubernetes ServiceAccounts by label.
+func DeleteServiceAccounts(
+	ctx context.Context,
+	kube *Kube,
+	namespace string,
+) error {
+	coreClient, err := kube.CoreV1ClientSet(namespace)
+	if err != nil {
+		return err
+	}
+	ServiceAccountList, err := coreClient.ServiceAccounts("").
+		List(ctx, metav1.ListOptions{LabelSelector: labelSelector})
+	for _, sa := range ServiceAccountList.Items {
+		err := coreClient.ServiceAccounts(sa.Namespace).
+			Delete(ctx, sa.Name, metav1.DeleteOptions{})
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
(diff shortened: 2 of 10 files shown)


installer/charts/_common/_helpers.tpl
```diff
@@ -42,6 +42,13 @@ app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
 {{- end }}
 
+{{/*
+Need deleted labels
+*/}}
+{{- define "common.needDeletedLabels" -}}
+delete: need-delete
```
Roming22: I think this label should be a bit more verbose. `rhtap-cli.redhat-appstudio.github.com//post-deploy: delete`
The helm definition should be renamed `common.postDeployDelete`.

pkg/k8s/delete_resources.go
```diff
@@ -0,0 +1,153 @@
...
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// labelSelector is the label set for resources to be deleted.
+const labelSelector = "delete=need-delete"
+
+// nsPrefix is the prefix for all namespaces rhtap installed.
+const nsPrefix = "rhtap"
```
Roming22: You cannot assume the namespace prefix.
Since we currently don't support:

- multiple RHTAP instances per cluster
- deploying charts in parallel

you can just get all resources with that selector and delete them without additional filtering.

pkg/k8s/delete_resources.go
```diff
@@ -0,0 +1,153 @@
...
+	for _, ns := range nsList.Items {
+		if strings.HasPrefix(ns.Name, nsPrefix) {
+			rhtapNamespaces = append(rhtapNamespaces, ns.Name)
+		}
+	}
+	return rhtapNamespaces, nil
+}
+
+// DeleteResources deletes temporary Kubernetes resources created during deployment.
+func DeleteResources(
```
Roming22, line 110: Isn't there some Roles/ClusterRoles/RoleBindings that are created as well?
Even if we don't have them now, it would be nice to handle those usecases, so we don't forget to add support if we need it later.

pkg/k8s/delete_resources.go
```diff
@@ -0,0 +1,153 @@
...
+	ctx context.Context,
+	kube *Kube,
+	namespace string,
+) error {
+	// Delete ClusterRoleBindings
```
Roming22, line 115: I'm surprised we have to go over each resource type individually.
Couldn't we use something like the CLI `kubectl get all --all-namespaces -l app=demo` command to list all resources that have the label, and loop on them to delete them?

pkg/k8s/delete_resources.go
```diff
@@ -0,0 +1,153 @@
...
+// RetryDeleteResources deletes temporary resources with retry 5 times.
+func RetryDeleteResources(
+	ctx context.Context,
+	kube *Kube,
+	namespace string,
+) error {
+	// Delete temporary resources
+	err := Retry(5, 10*time.Second, func() error {
+		err := DeleteResources(ctx, kube, namespace)
+		if err != nil {
```
Roming22: Looks like this can be simplified to `return err`.

pkg/k8s/delete_resources.go
```diff
@@ -0,0 +1,153 @@
...
+			return err
+		}
+		return nil
+	})
+	if err != nil {
```
Roming22: Looks like this can be simplified to `return err`.

pkg/subcmd/deploy.go
```diff
@@ -122,25 +122,37 @@ func (d *Deploy) Run() error {
 
 		err := i.SetValues(d.cmd.Context(), &d.cfg.Installer, string(valuesTmpl))
 		if err != nil {
+			// Delete temporary resources
```
Roming22: Why is it needed here? I don't think any resources have been applied to the cluster at that time.

pkg/subcmd/deploy.go
```diff
@@ -122,25 +122,37 @@ func (d *Deploy) Run() error {
...
+			_ = k8s.RetryDeleteResources(d.cmd.Context(), d.kube, d.cfg.Installer.Namespace)
 			return err
 		}
 		if d.flags.Debug {
 			i.PrintRawValues()
 		}
 
 		if err := i.RenderValues(); err != nil {
+			// Delete temporary resources
```
Roming22: Why is it needed here? I don't think any resources have been applied to the cluster at that time.

pkg/subcmd/deploy.go
```diff
@@ -122,25 +122,37 @@ func (d *Deploy) Run() error {
...
 
 		if err = i.Install(d.cmd.Context()); err != nil {
+			// Delete temporary resources
+			_ = k8s.RetryDeleteResources(d.cmd.Context(), d.kube, d.cfg.Installer.Namespace)
 			return err
 		}
 		fmt.Printf("############################################################\n\n")
 	}
 
+	// Delete temporary resources
```
Roming22: You can remove this if you update the code above to:
```
		err = i.Install(d.cmd.Context());
		// Delete temporary resources
		_ = k8s.RetryDeleteResources(d.cmd.Context(), d.kube, d.cfg.Installer.Namespace) 
		if err != nil {
			return err
		}
```

pkg/subcmd/deploy.go
lingyzhuang: It's here to delete resources that are created in previous charts.

pkg/subcmd/deploy.go
lingyzhuang: It's here to delete resources that are created in previous charts.

pkg/subcmd/deploy.go
Roming22: I feel it's unnecessary. You might want to do it once before the for-loop, to handle the case where an install was interrupted, but within the loop you only need to do it once at the end.

pkg/subcmd/deploy.go
Roming22: See my previous comment.

pkg/subcmd/deploy.go
lingyzhuang: I'll use an example to explain my thoughts:
If installation of `[10/12] 'charts/rhtap-app-namespaces'` failed at `i.SetValues` or `i.RenderValues`, it will return err immediately, the resources created by previous steps(e.g. `charts/rhtap-pipelines`) will not be deleted as the delete operations will be done only in `i.Install`.
If add the delete call before the for-loop, the resources will not be deleted until the beginning of the next installation.
Is that as expected?

pkg/k8s/delete_resources.go
lingyzhuang: Added Roles/ClusterRoles/RoleBindings deletion.

pkg/k8s/delete_resources.go
lingyzhuang: Use DeleteColletion instead

installer/charts/_common/_helpers.tpl
```diff
@@ -42,6 +42,13 @@ app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
 {{- end }}
 
+{{/*
+Need deleted labels
+*/}}
+{{- define "common.needDeletedLabels" -}}
```
Roming22: You should rename it `common.postDeployDeleteLabels`

Roming22 (commented):
Sorry I forgot to publish my review...

installer/charts/_common/_helpers.tpl
lingyzhuang: Done

Roming22 (commented):
Looks good to me.
Waiting on feedback from Otavio regarding the minIO SAs, and code review.

pkg/k8s/kube_fake.go
```diff
@@ -88,6 +89,16 @@ func (f *FakeKube) GetDynamicClientForObjectRef(
 	return dynamicClient.Resource(gvr), nil
 }
 
+func (f *FakeKube) RbacV1ClientSet(
```
otaviof: nit: The naming should be `RBACV1ClientSet` instead.

pkg/subcmd/deploy.go
```diff
@@ -135,7 +135,10 @@ func (d *Deploy) Run() error {
 			i.PrintValues()
 		}
 
-		if err = i.Install(d.cmd.Context()); err != nil {
+		err = i.Install(d.cmd.Context())
+		// Delete temporary resources
+		_ = k8s.RetryDeleteResources(d.cmd.Context(), d.kube, d.cfg.Installer.Namespace)
```
otaviof: We should capture and handle the possible errors, a alternative is making sure we print these errors when debug logging.

otaviof (commented):
Just a couple of comments.

pkg/subcmd/deploy.go
lingyzhuang: Updated

Roming22 (approved):
/lgtm

PR #627, opened 2025-04-08
Fix snyk Task and remove PVC

lcarva:
The build pipeline uses the Trusted Artifact[1] pattern. This means that OCI artifacts are used to share data between Tasks instead of a PVC, including the cloned source code.

The snyk task previously used did not support TA and looked for the source code to be scanned within an empty PVC. No scan was performed.

This commit uses the TA version of the snyk task. As a cleanup, it also removes the remaining references to the PVC. This should make the build pipeline lighter as it needs less resources.

Ref: https://issues.redhat.com/browse/RHTAP-4520

[1] https://konflux-ci.dev/architecture/ADR/0036-trusted-artifacts.html

### DIFF
--- a/.tekton/rhtap-cli-pull-request.yaml
+++ b/.tekton/rhtap-cli-pull-request.yaml
@@ -64,9 +64,6 @@ spec:
         - name: kind
           value: task
         resolver: bundles
-      workspaces:
-      - name: workspace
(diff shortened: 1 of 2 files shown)


Roming22 (approved):
/lgtm

PR #646, opened 2025-04-10
Hide the Azure pull request tab [1.5]

lcarva:
The default config for the azure frontend plugin enables both EntityAzurePullRequestsContent and EntityAzurePipelinesContent. The first is not supported by RHTAP since Azure is not a git provider option. Both are enabled by the `dev.azure.com/project-repo` Component annotation. As such, there is no possible combination of Component annotations that would allow us to enable EntityAzurePipelinesContent but disable EntityAzurePullRequestsContent.

This commit overwrites the default config for the azure plugin to omit EntityAzurePullRequestsContent.


Roming22 (approved):
/lgtm

PR #698, opened 2025-04-23
Migrating quay storage from Minio to ODF (Noobaa)

dperaza4dustbit:
Implements [RHTAP-4879](https://issues.redhat.com//browse/RHTAP-4879)

Includes the installation of Red Hat OpenShift Data Foundation ODF (previously known as OpenShift Container Storage OCS)

### DIFF
--- a/installer/charts/rhtap-infrastructure/templates/odf/backing-store.yaml
+++ b/installer/charts/rhtap-infrastructure/templates/odf/backing-store.yaml
@@ -1,0 +1,23 @@
+{{- if .Values.infrastructure.odf.enabled }}
+apiVersion: noobaa.io/v1alpha1
+kind: BackingStore
+metadata:
+  finalizers:
+  - noobaa.io/finalizer
+  labels:
+    app: noobaa
+  name: {{ .Values.infrastructure.odf.backingStoreName }}
+  namespace: {{ .Values.infrastructure.odf.namespace }}
+spec:
+  pvPool:
+    numVolumes: 1
+    resources:
+      limits:
+        cpu: 1000m
+        memory: 4000Mi
+      requests:
+        cpu: 800m
+        memory: 800Mi
+        storage: {{ .Values.infrastructure.odf.backingStorageSize }}
+  type: pv-pool
+{{- end }}
--- a/installer/charts/rhtap-infrastructure/templates/odf/enable-odf-console-plugin.yaml
+++ b/installer/charts/rhtap-infrastructure/templates/odf/enable-odf-console-plugin.yaml
@@ -1,0 +1,30 @@
+{{- if .Values.infrastructure.odf.enabled }}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: patch-console-enable-odf
+  namespace: {{ .Values.infrastructure.odf.namespace }}
+  annotations:
+    "helm.sh/hook": post-install,post-upgrade
+    "helm.sh/hook-delete-policy": hook-succeeded
+spec:
+  backoffLimit: 3
+  template:
+    spec:
+      serviceAccountName: patch-console-sa
+      restartPolicy: OnFailure
+      containers:
+        - name: patch-console
+          image: quay.io/openshift/origin-cli:latest
+          command:
+            - /bin/bash
+            - -c
+            - |
+              echo "Patching Console resource to enable odf-console plugin..."
+              if oc get console.operator.openshift.io cluster -o json | grep -q '"odf-console"'; then
+                echo "odf-console plugin already enabled. Skipping patch."
+              else
+                oc patch console.operator.openshift.io cluster --type=json -p '[{"op":"add","path":"/spec/plugins/-","value":"odf-console"}]'
+                echo "Console plugin enabled"
+              fi
+{{- end }}
--- a/installer/charts/rhtap-infrastructure/templates/odf/patch-default-bucketclass.yaml
+++ b/installer/charts/rhtap-infrastructure/templates/odf/patch-default-bucketclass.yaml
@@ -1,0 +1,41 @@
+{{- if .Values.infrastructure.odf.enabled }}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: patch-noobaa-default-bucketclass
+  namespace: {{ .Values.infrastructure.odf.namespace }}
+  annotations:
+    "helm.sh/hook": post-install,post-upgrade
+    "helm.sh/hook-delete-policy": hook-succeeded
+spec:
+  template:
+    backoffLimit: 3
+    spec:
+      serviceAccountName: patch-bucketclass-sa
+      restartPolicy: OnFailure
+      containers:
+        - name: patch-bucketclass
+          image: quay.io/openshift/origin-cli:latest
+          command:
+            - /bin/bash
+            - -c
+            - |
+              set -e
+              echo "Waiting for noobaa-default-bucket-class to be ready..."
+              for i in {1..30}; do
+                if oc get bucketclass noobaa-default-bucket-class -n {{ .Values.infrastructure.odf.namespace }} > /dev/null 2>&1; then
+                  echo "BucketClass found!"
+                  break
+                fi
+                echo "Waiting for BucketClass... ($i/30)"
+                sleep 10
+              done
+
+              echo "Patching noobaa-default-bucket-class to use backing store '{{ .Values.infrastructure.odf.backingStoreName }}'..."
+              
+              oc patch bucketclass noobaa-default-bucket-class -n {{ .Values.infrastructure.odf.namespace }} --type=json -p="[
+                {\"op\": \"replace\", \"path\": \"/spec/placementPolicy/tiers/0/backingStores\", \"value\": [\"{{ .Values.infrastructure.odf.backingStoreName }}\"]}
+              ]"
+
+              echo "Patch complete."
+{{- end }}
--- a/installer/charts/rhtap-infrastructure/templates/odf/post-install-hooks-rbac.yaml
+++ b/installer/charts/rhtap-infrastructure/templates/odf/post-install-hooks-rbac.yaml
@@ -1,0 +1,65 @@
+{{- if .Values.infrastructure.odf.enabled }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: patch-console-sa
+  namespace: {{ .Values.infrastructure.odf.namespace }}
+  labels: {{- include "common.postDeployDeleteLabels" . | nindent 4 }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: patch-console-role
+  labels: {{- include "common.postDeployDeleteLabels" . | nindent 4 }}
+rules:
+  - apiGroups: ["operator.openshift.io"]
+    resources: ["consoles"]
+    verbs: ["get", "patch"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: patch-console-rolebinding
+  labels: {{- include "common.postDeployDeleteLabels" . | nindent 4 }}
+subjects:
+  - kind: ServiceAccount
+    name: patch-console-sa
+    namespace: {{ .Values.infrastructure.odf.namespace }}
+roleRef:
+  kind: ClusterRole
+  name: patch-console-role
+  apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: patch-bucketclass-sa
+  namespace: {{ .Values.infrastructure.odf.namespace }}
+  labels: {{- include "common.postDeployDeleteLabels" . | nindent 4 }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: patch-bucketclass-role
+  namespace: {{ .Values.infrastructure.odf.namespace }}
+  labels: {{- include "common.postDeployDeleteLabels" . | nindent 4 }}
+rules:
+  - apiGroups: ["noobaa.io"]
+    resources: ["bucketclasses"]
+    verbs: ["get", "patch"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: patch-bucketclass-rolebinding
+  namespace: {{ .Values.infrastructure.odf.namespace }}
+  labels: {{- include "common.postDeployDeleteLabels" . | nindent 4 }}
+subjects:
+  - kind: ServiceAccount
+    name: patch-bucketclass-sa
+    namespace: {{ .Values.infrastructure.odf.namespace }}
+roleRef:
+  kind: Role
+  name: patch-bucketclass-role
+  apiGroup: rbac.authorization.k8s.io
+{{- end }}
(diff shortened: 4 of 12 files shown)


dperaza4dustbit:
Installed ODF manually and configured it so that quay can manage its own storage via ObejectBucketClaim API

dperaza4dustbit:
Leaving in draft until I do a bit more testing

installer/charts/values.yaml.tpl
```diff
@@ -86,6 +87,9 @@ subscriptions:
   redHatQuay:
     enabled: {{ $quay.Enabled }}
     managed: {{ and $quay.Enabled $quay.Properties.manageSubscription }}
+  redHatOpenShiftDataFoundation:
+    enabled: {{ $odfEnabled }}
+    managed: {{ $odfEnabled }}
```
Roming22, line 98: I think you need to control the subscription from the `config.yaml`, which means adding a new feature.

installer/charts/values.yaml.tpl
```diff
@@ -94,6 +98,16 @@ subscriptions:
 minIOOperator:
   enabled: {{ $minIOOperatorEnabled }}
 
+
+#
```
Roming22: We use this comment to document the chart that uses the value. I would argue that maybe you should create an `rhtap-odf` chart instead of adding the config to `rhtap-infrastructure`.

dperaza4dustbit:
First round of testing of PR looking good. Realized I'm missing logic to set default baking store for the noobaa default storage class. Also realized need to figure out how to enable the console plugin for ODF subscription. Enabled console plugin and default backing store manually and quay storage is working well.

installer/charts/values.yaml.tpl
dperaza4dustbit: This was first step as we discussed already, and I thought to add a new feature but I realized that I would not want the user to control weather to enable or not directly. That is why I set odfEnabled base on quay and tpa enabled flags. So.. not sure what to do, cause a feature implies you control enablement.

installer/charts/values.yaml.tpl
dperaza4dustbit: I like the fact that I can control the resources we need for ODF without having to disable or enable a whole chart, and there is only two resources we need to generate. So not sure I agree, but could go either way if we think we will need to follow a different pattern we follow with minio today. The minio operator is getting installed via a helm chart vs odf via subscription

installer/charts/rhtap-infrastructure/templates/odf/backing-store.yaml
```diff
@@ -0,0 +1,24 @@
...
+    numVolumes: 1
+    resources:
+      limits:
+        cpu: 1000m
+        memory: 4000Mi
+      requests:
+        cpu: 800m
+        memory: 800Mi
+        storage: {{ .Values.odf.backingStorageSize }}
+    storageClass: gp3-csi
```
dperaza4dustbit: Need to remove this so that it picks up the default storage class, this happens to be the default, so functionally is working as expected

installer/charts/values.yaml.tpl
dperaza4dustbit: Is it possible to exposed properties for a dependency like in this case I want to expose in config.yaml the bakingStoreSize, but not whether to enable or not ods, this is controlled indirectly by the drivers that need odf (quay and tpa)

dperaza4dustbit: Maybe the best way to do this is to expose property for required object store size for both TPA and Quay and then do some math in values.yamp.tpl to calculate overall space required. I suggest we do this when we implement TPA

installer/charts/rhtap-openshift/templates/projects.yaml
```diff
@@ -6,16 +6,21 @@
...
-        (not (lookup "project.openshift.io/v1" "Project" "" $p ))
-  }}
+  {{- if not (lookup "project.openshift.io/v1" "Project" "" $p ) }}
+    {{- if hasPrefix "openshift-" $p }}
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: {{ $p }}
+    {{- else }}
```
dperaza4dustbit: Need this cause project request will reject an namespace that starts with openshift-

installer/charts/values.yaml.tpl
```diff
@@ -86,6 +91,14 @@ subscriptions:
   redHatQuay:
     enabled: {{ $quay.Enabled }}
     managed: {{ and $quay.Enabled $quay.Properties.manageSubscription }}
+  redHatOpenShiftDataFoundation:
+    enabled: {{ $odfEnabled }}
+    managed: {{ $odfEnabled }}
+    namespace: {{ $odfNamespace }}
+    channel: stable-4.18
```
dperaza4dustbit: need to figure out how to make this dynamic depending on current openshift version. Take suggestions in the meantime

installer/charts/rhtap-infrastructure/templates/odf/backing-store.yaml
dperaza4dustbit: This is done and verified

dperaza4dustbit:
removing draft as I think this ready, only missing how to dynamically figure out version of current openshift so that we don't have to hardcode ODF channel in values.yaml.tpl

installer/charts/rhtap-openshift/templates/projects.yaml
dperaza4dustbit: Something is causing the openshift-sorage to be recreated when helm upgrade happens. Lets hold on merging this PR for now, unit I can verify why.

dperaza4dustbit: I know what the problem is, fixing

dperaza4dustbit (commented):
Hold on merging this fixing a problem with rhtap-openshift chart

installer/charts/values.yaml.tpl
Roming22: Object store size is going to be fixed at the moment, and should be set in `values.yaml.tpl`. Making it dynamic is a discussion for the "t-shirt size" feature.

installer/charts/values.yaml.tpl
Roming22: I'm fine with the subscription being enabled in `values.yaml.tpl`.

installer/charts/rhtap-openshift/templates/projects.yaml
dperaza4dustbit: This is fixed

installer/charts/values.yaml.tpl
dperaza4dustbit: Got it

dperaza4dustbit:
> Hold on merging this fixing a problem with rhtap-openshift chart

This is fix

installer/charts/values.yaml.tpl
dperaza4dustbit: Have a path forward with some changes to variables we fit to the rendering on values.yaml.tpl, working on this now. If this merges before I'm done will send another PR.

installer/charts/rhtap-infrastructure/templates/odf/enable-odf-console-plugin.yaml
```diff
@@ -0,0 +1,28 @@
+{{- if .Values.odf.enabled }}
+apiVersion: batch/v1
+kind: Job
```
Roming22, line 3: I've found that Jobs tend to cause issues. Can you switch to a Pod, and make sure it is idempotent?

installer/charts/rhtap-infrastructure/templates/odf/patch-default-bucketclass.yaml
```diff
@@ -0,0 +1,40 @@
+{{- if .Values.odf.enabled }}
+apiVersion: batch/v1
+kind: Job
```
Roming22, line 3: See my previous comment about Jobs.

installer/charts/rhtap-infrastructure/templates/odf/enable-odf-console-plugin.yaml
```diff
@@ -0,0 +1,28 @@
...
+          command:
+            - /bin/bash
+            - -c
+            - |
+              echo "Patching Console resource to enable odf-console plugin..."
+              if oc get console.operator.openshift.io cluster -o json | grep -q '"odf-console"'; then
+                echo "odf-console plugin already enabled. Skipping patch."
+              else
+                oc patch console.operator.openshift.io cluster --type=json -p '[{"op":"add","path":"/spec/plugins/-","value":"odf-console"}]'
+              fi
```
Roming22, line 29: Add a final 'echo' statement so it's easy to see the success status from looking at the logs (eg `echo "# Success"`).

installer/charts/rhtap-infrastructure/templates/odf/post-install-hooks-rbac.yaml
```diff
@@ -0,0 +1,59 @@
+{{- if .Values.odf.enabled }}
```
Roming22: If those service accounts and associated resources are useless once the deployment is complete, please add the [required annotations](https://github.com/redhat-appstudio/rhtap-cli/blob/529d4203f10b1230cb4f699c705e392a06786e0d/installer/charts/_common/_service-account.tpl#L7) so they get cleaned up.

installer/charts/rhtap-openshift/templates/projects.yaml
```diff
@@ -15,6 +15,13 @@ apiVersion: project.openshift.io/v1
 kind: ProjectRequest
 description: {{ $p }}
 displayName: {{ $p }}
+metadata:
```
Roming22, line 18: I'm fine with merging that if it works, but I thought creating `openshift-*` namespaces/projects was sometimes messing up with some operators deployment.

installer/charts/values.yaml.tpl
Roming22: We can merge as is for the moment, but the issue that if the subscription for ODF is already on the cluster Helm is going to complain about it.

installer/charts/values.yaml.tpl
```diff
@@ -10,6 +10,9 @@
 {{- $ingressDomain := required "OpenShift ingress domain" .OpenShift.Ingress.Domain -}}
 {{- $ingressRouterCA := required "OpenShift RouterCA" .OpenShift.Ingress.RouterCA -}}
 {{- $minIOOperatorEnabled := or $tpa.Enabled $quay.Enabled -}}
+{{- $odfEnabled := or $tpa.Enabled $quay.Enabled -}}
+{{- $odfNamespace := "openshift-storage" -}}
```
Roming22, line 14: Maybe the namespace/channel definition could be moved closer to where they're being used.

installer/charts/values.yaml.tpl
Roming22: The comment about moving this to rhtap-infrastructure (since it's the chart that uses those values) still stands though.

installer/charts/rhtap-openshift/templates/projects.yaml
dperaza4dustbit: It does not work with Project Request cause we have some policies in OpenShift to block, but should be fine when we create as a namespace. This is required cause openshift-storage Project is create by ODF if done from UI but not when using subscription CR. This is working fine at the moment.

installer/charts/rhtap-infrastructure/templates/odf/enable-odf-console-plugin.yaml
dperaza4dustbit: Hum... right now we are using more Jobs than plain pots for post-install hooks. Are we planning to clean that up? What problems have you seen? This is one of the reasons jobs exist. https://github.com/search?q=repo%3Aredhat-appstudio%2Frhtap-cli%20post-install&type=code. Also, they provide an example in helm docs with Job https://helm.sh/docs/topics/charts_hooks/

installer/charts/rhtap-infrastructure/templates/odf/enable-odf-console-plugin.yaml
dperaza4dustbit: Sure thing

installer/charts/rhtap-infrastructure/templates/odf/post-install-hooks-rbac.yaml
dperaza4dustbit: Good Catch

installer/charts/values.yaml.tpl
dperaza4dustbit: Sure thing, makes sense

installer/charts/values.yaml.tpl
dperaza4dustbit: Isn't the error from helm something we want, I want to make sure that when I decide to manage subscription through installer subscription CR is only managed by helm

installer/charts/values.yaml.tpl
dperaza4dustbit: I still don't see the benefit of an additional chart. We are not doing that with minio storage, why do we want to do it with ODF storage. It is just infra dependency.

installer/charts/values.yaml.tpl
Roming22: Let's have a follow up story in RHTAP-4855 to automate the usecase of detecting if a subscription is already present.

dperaza4dustbit: Sounds good

installer/charts/values.yaml.tpl
dperaza4dustbit: Understand now the ask here, thanks for time on the call to clarify. Moving the odf vars to infrastructure object

installer/charts/rhtap-infrastructure/templates/odf/enable-odf-console-plugin.yaml
dperaza4dustbit: Thanks for time in the call to clarify, just letting others know we decided to leave jobs for now with some limited backoffLimit to avoid long time to fail with 6 being the default

installer/charts/values.yaml.tpl
dperaza4dustbit: Since namespace and enabled is use in openshift, subscription and infrastructure leaving here. Moving channel to be under subscription section.

installer/charts/rhtap-infrastructure/templates/odf/enable-odf-console-plugin.yaml
dperaza4dustbit: Done

installer/charts/values.yaml.tpl
dperaza4dustbit: done

installer/charts/values.yaml.tpl
dperaza4dustbit: Done, create story: https://issues.redhat.com/browse/RHTAP-4905

installer/charts/values.yaml.tpl
dperaza4dustbit: Done

installer/charts/rhtap-infrastructure/templates/odf/post-install-hooks-rbac.yaml
dperaza4dustbit: Done

Roming22 (approved):
/lgtm

PR #716, opened 2025-04-29
fix: improve GH App integration webpage

Roming22:
Add a button to make it more obvious that an action is required.


dperaza4dustbit (approved):
/lgtm

PR #719, opened 2025-05-02
fix: add /licenses folder to image

Roming22:
Licensing is checked by the 'ecosystem-cert-preflight-check' task. Though it is currently just a warning, adding the licensing information will prevent a failure if the check becomes mandatory.


dperaza4dustbit (approved):
/lgtm

PR #739, opened 2025-05-07
Decrease the chance of CI failure due to TektonConfig cr being there

dperaza4dustbit:
The logic in the tekton [controller](https://github.com/tektoncd/operator/blob/main/pkg/reconciler/shared/tektonconfig/controller.go#L127) code  which calls the [ensureInstance](https://github.com/tektoncd/operator/blob/main/pkg/reconciler/shared/tektonconfig/instance.go#L64) function suggest that it is best effort to create a config instance if is not there already. So my theory is that even if the AUTOINSTALL_COMPONENTS env var is set to false there could still be a chance of a config being created and that is why we see considerable runs of install failing becase helm cannot take ownership of the resource.

I think this could improve our chances of success with our CI.

### DIFF
--- a/installer/charts/rhtap-subscriptions/templates/openshift-pipelines/delete_tekton_config.yaml
+++ b/installer/charts/rhtap-subscriptions/templates/openshift-pipelines/delete_tekton_config.yaml
@@ -1,0 +1,35 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: delete-tekton-config-if-exists
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/hook": post-install
+    "helm.sh/hook-delete-policy": hook-succeeded
+spec:
+  backoffLimit: 8
+  template:
+    spec:
+      serviceAccountName: delete-tekton-config-sa
+      restartPolicy: OnFailure
+      containers:
+        - name: delete-tekton-config
+          image: quay.io/openshift/origin-cli:latest
+          resources:
+            limits:
+              memory: 128Mi
+          command:
+            - /bin/bash
+            - -c
+            - |
+              if [[ "{{ .Values.subscriptions.openshiftPipelines.managed }}" == "true" ]]; then
+                  echo "Deleting TektonConfig resource if it exists to allow helm to own the resource"
+                  oc delete tektonconfig config --ignore-not-found
+              else
+                  echo "Skipping TektonConfig resource delete as pipeline subscription is not managed"
+              fi
+              if [[ $? -ne 0 ]]; then
+                  echo "Sleeping for 5 seconds before letting job restart pod"
+                  sleep 5
+                  exit 1
+              fi
--- a/installer/charts/rhtap-subscriptions/templates/openshift-pipelines/post-install-hook-rbac.yaml
+++ b/installer/charts/rhtap-subscriptions/templates/openshift-pipelines/post-install-hook-rbac.yaml
@@ -1,0 +1,30 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: delete-tekton-config-sa
+  namespace: {{ .Release.Namespace }}
+  labels: {{- include "common.postDeployDeleteLabels" . | nindent 4 }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: delete-tekton-config-role
+  labels: {{- include "common.postDeployDeleteLabels" . | nindent 4 }}
+rules:
+  - apiGroups: ["operator.tekton.dev"]
+    resources: ["tektonconfigs"]
+    verbs: ["get", "delete"]
+---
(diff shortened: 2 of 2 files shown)


installer/charts/rhtap-subscriptions/templates/delete_tekton_config.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: delete-tekton-config-if-exists
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/hook": post-install
+    "helm.sh/hook-delete-policy": hook-succeeded
+spec:
```
Roming22: There's a couple of issues raised by SonarCloud.

installer/charts/rhtap-subscriptions/templates/delete_tekton_config.yaml
```diff
@@ -0,0 +1,23 @@
...
+      restartPolicy: OnFailure
+      containers:
+        - name: delete-tekton-config
+          image: quay.io/openshift/origin-cli:latest
+          command:
+            - /bin/bash
+            - -c
+            - |
+              echo "Deleting TektonConfig config resource if it exist to allow helm to own the resource"
+              oc delete tektonconfig config --ignore-not-found
```
Roming22: If feel that the TektonConfig should be deleted only if the CLI is managing the subscription and the resource is not annotated as belonging to a helm deployment.

Roming22 (changes requested):

installer/charts/rhtap-subscriptions/templates/delete_tekton_config.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: delete-tekton-config-if-exists
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/hook": post-install
```
Roming22: I feel it should be a `pre-install` hook of the `tssc-infrastructure` chart, and the job/SA should be in `installer/charts/tssc-infrastructure/templates/openshift-pipelines/tektonconfig/`.

dperaza4dustbit: I thought the same, and I tried it, but pre-install hooks run before the templates are there and it will be missing the RBAC to be able to delete.

installer/charts/rhtap-subscriptions/templates/delete_tekton_config.yaml
dperaza4dustbit: Agree with checking if we just installed the subscription. But it won't be a helm deployment cause will only run on install not upgrades. Once install is done, helm should own the CR

installer/charts/rhtap-subscriptions/templates/delete_tekton_config.yaml
dperaza4dustbit: I saw that, will take care of those

installer/charts/rhtap-subscriptions/templates/delete_tekton_config.yaml
Roming22: Understood. Can you at least move the files into an `openshift-pipelines` subdir?

dperaza4dustbit: sounds good

dperaza4dustbit: done

installer/charts/rhtap-subscriptions/templates/openshift-pipelines/delete_tekton_config.yaml
```diff
@@ -0,0 +1,30 @@
...
+  name: delete-tekton-config-if-exists
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/hook": post-install
+    "helm.sh/hook-delete-policy": hook-succeeded
+spec:
+  backoffLimit: 3
+  template:
+    spec:
+      serviceAccountName: delete-tekton-config-sa
```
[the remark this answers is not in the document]
dperaza4dustbit: This is required to be able to run oc commands under pod as an sa

installer/charts/rhtap-subscriptions/templates/delete_tekton_config.yaml
dperaza4dustbit: Done

installer/charts/rhtap-subscriptions/templates/delete_tekton_config.yaml
dperaza4dustbit: done

dperaza4dustbit:
Tested multiple times on my dev environment I feel this is ready, removing from draft

Roming22 (approved):
/lgtm

Roming22 (approved):

PR #745, opened 2025-05-08
fix: ignore deleted repos in GitLab

Roming22:


Roming22:
Force merging as there's no impact on the CI.

PR #763, opened 2025-05-12
RHTAP-2871: Update rhtap-install task to set DH catalog url from tssc-sample-templates repo

chetna14manku:
This change is required for testing tssc-sample-templates PR changes in Konflux.


xinredhat (approved):
/lgtm

PR #764, opened 2025-05-12
RHTAP-4928 Add pict file parameter in e2e-pipeline

jkopriva:


prietyc123:
@jkopriva I am wondering why not use randomization for periodic job pict configuration. We can target to run first 5 random configuration each time we run periodic job WDYT? This would ensure each time we will run 5 different pict configuration

jkopriva:
@prietyc123 I was talking about it with Radim and he also suggested this approach - the PR is updated to to run 4 jobs randomly based on the periodic-all.pict file if the pict file path is set, otherwise it is using default value. PTAL and comment. thank you!

integration-tests/pipelines/e2e-main-pipeline.yaml
```diff
@@ -69,12 +77,20 @@ spec:
...
+                value: $(params.PICT_FILE_PATH)
             script: |
               #!/usr/bin/env bash
               set -x
               base64 -d <<< "$(tasks.get-pict-file.results.pict-file)" > main.pict
               cat main.pict
-              pict ./main.pict -o:1 -f:json > $(results.configs-json.path)
+              if [[ $PICT_FILE_LOCATION  =~ "periodic" ]]; then
+                ## Generate pict file for periodic job
+                pict ./main.pict -r -o:1 -f:json > $(results.configs-json.path)
```
prietyc123, line 90: Randomisation looks good +1 , but could you please help me out how do we ensure only 4 pict configuration runs? Am I missing something? 
IMO it should be controlled [here](https://github.com/redhat-appstudio/rhtap-cli/pull/764/files#diff-8b3a6d633efd750bda2a47658f938f3cecd65d6b4ca95be00405736f34488dd9R172), no?

integration-tests/pict-models/periodic-jobs/README.md
```diff
@@ -0,0 +1,2 @@
+## Pict files for periodic jobs
+Pict files are base on default.pict file (all constrains are copied from it)
```
prietyc123:
```suggestion
Pict files are based on default.pict file (all constraints are copied from it)
```

prietyc123 (dismissed):

integration-tests/pipelines/e2e-main-pipeline.yaml
jkopriva: it is achivied by this parameter: `-o:1` - by default it is 2 - so it generates more configs, but we are changing it also for pr checks to 1, so it will generate 4 configs

prietyc123:
@jkopriva Could you please squash all the commit. Changes looks good to me 👍

PR #765, opened 2025-05-12
fix: temporary fix for Postgres operator

Roming22:
There is currenlty an issue with the marketplace operator. Switching to the community version until the issue is solved.


Roming22:
Closed in favor of #771

PR #787, opened 2025-05-15
fix: fix config.yaml path in deploy.sh

Roming22:


Roming22:
Force merging as there is no impact on the CI.

PR #790, opened 2025-05-16
RHTAP-4553: `tssc config` Subcommand vs. Global Flags (`--dry-run`)

otaviof:
Including the application's global flags reference to identify when `--dry-run` flag is informed on `tssc config` subcommand.

### DIFF
--- a/pkg/subcmd/config.go
+++ b/pkg/subcmd/config.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/redhat-appstudio/rhtap-cli/pkg/chartfs"
 	"github.com/redhat-appstudio/rhtap-cli/pkg/config"
+	"github.com/redhat-appstudio/rhtap-cli/pkg/flags"
 	"github.com/redhat-appstudio/rhtap-cli/pkg/k8s"
 
 	"github.com/spf13/cobra"
@@ -16,6 +17,7 @@ import (
 type Config struct {
 	cmd    *cobra.Command   // cobra command
 	logger *slog.Logger     // application logger
+	flags  *flags.Flags     // global flags
 	cfs    *chartfs.ChartFS // embedded filesystem
 	kube   *k8s.Kube        // kubernetes client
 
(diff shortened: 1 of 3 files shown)


pkg/subcmd/config.go
```diff
@@ -169,6 +177,32 @@ func (c *Config) runCreate() error {
 	return err
 }
 
+// runDelete controls the deletion process.
+func (c *Config) runDelete() error {
+	if c.flags.DryRun {
```
dperaza4dustbit, line 188: What about say It would delete CM config in namespace RHTAP or something like that, basically tell the user what it would do if dry run was not enabled

pkg/subcmd/config.go
```diff
@@ -169,6 +177,32 @@ func (c *Config) runCreate() error {
...
+}
+
+// runGet controls the cluster configuration retrieval process.
+func (c *Config) runGet() error {
+	c.log().Debug("Retrieving the cluster configuration")
+	cfg, err := c.manager.GetConfig(c.cmd.Context())
+	if err != nil {
+		if c.create && c.flags.DryRun {
+			c.log().Warn(
+				"[DRY-RUN] Configuration does not exist in the cluster, yet.")
```
dperaza4dustbit, line 207: Same comment here: I would just print out what it would try to do if dry run is true

Roming22: I'm not sure `--dry-run` is relevant on a `get`, since it's a read-only operation.

dperaza4dustbit: yeah maybe just do the get

otaviof: Well, the tool supports using `--get` and `--create` at the same time, and that's something I use all the time. Otherwise, `--create --get --dry-run` would return error, something we should avoid when `--dry-run` is informed.

pkg/subcmd/config.go
otaviof: Sure! Based on what we've spoken, please consider the latest changes.

dperaza4dustbit: looks good to me

dperaza4dustbit (approved):
/lgtm

PR #792, opened 2025-05-16
feat: tssc renaming

Roming22:
Update docs, descriptions, messages, etc.

### DIFF
--- a/pkg/k8s/openshift.go
+++ b/pkg/k8s/openshift.go
@@ -7,12 +7,12 @@ import (
 	"log/slog"
 	"time"
 
+	configv1 "github.com/openshift/api/config/v1"
 	v1 "github.com/openshift/api/operator/v1"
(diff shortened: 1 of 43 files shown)


pkg/k8s/openshift.go
```diff
@@ -7,12 +7,12 @@ import (
 	"log/slog"
 	"time"
 
+	configv1 "github.com/openshift/api/config/v1"
```
dperaza4dustbit, line 10: nothing to do with rename, but it looks harmless moving dependencies around

dperaza4dustbit (approved):
/lgtm

lingyzhuang (approved):

pkg/k8s/openshift.go
Roming22: That's the IDE auto format doing its thing...

PR #812, opened 2025-05-19
fix: TektonConfig workaround was broken by tssc renaming

Roming22:


Roming22:
Force merging as discussed with @otaviof

PR #779, opened 2025-05-14
Upgrade TPA to 2.0

lingyzhuang:
Upgrade TPA to 2.0.0

- Minio or ODF are not used for storage control. Set `type` to `filesystem` here.

- API version for TPA is `v2`, so this PR should be merged with: https://github.com/redhat-appstudio/tssc-sample-pipelines/pull/85

Jira: https://issues.redhat.com/browse/RHTAP-4854

PR #824, opened 2025-05-28
RHTAP-5179: Capture Curl Return Code (SC2181)

otaviof:
Fixing [SC2181](https://www.shellcheck.net/wiki/SC2181) for the URL test script.

### DIFF
--- a/installer/scripts/test-url.sh
+++ b/installer/scripts/test-url.sh
@@ -29,12 +29,15 @@ probe_url() {
             --show-error \
             --fail \
             --location \
+            --insecure \
+            --max-time 30 \
             --output /dev/null \
             --write-out "%{http_code}" \
             "${URL}"
-    )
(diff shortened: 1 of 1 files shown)


dperaza4dustbit (approved):
/lgtm

installer/scripts/test-url.sh
lingyzhuang, line 1: I got `curl: (60) SSL certificate problem: self-signed certificate in certificate chain` while installing TPA, so maybe `-k` is needed for the `curl` command.

otaviof: Sure, that's a good idea, thanks! Please consider the latest changes.

otaviof: I'm also adding `--max-time` to restrict the amount of time for each request.

Roming22:
Included in #779.

PR #826, opened 2025-05-29
fix: revert 'cli' to 'walker' in keycloak

Roming22:


otaviof (approved):
/lgtm

installer/charts/tssc-tpa/values.yaml
```diff
@@ -96,13 +96,13 @@ redhat-trusted-profile-analyzer:
         clientId: frontend
       # A specialized crawler to interact with external data sources, this client
       # is used by other TPA modules.
-      cli:
-        clientId: cli
+      walker:
```
lingyzhuang: Set it to `cli`

installer/charts/values.yaml.tpl
```diff
@@ -369,12 +369,12 @@ redhat-trusted-profile-analyzer:
   oidc: &tpaOIDC
     issuerUrl: {{ $tpaOIDCIssuerURL }}
     clients:
-      cli:
+      walker:
```
lingyzhuang: Set it to `cli`

installer/charts/tssc-tpa-realm/values.yaml
```diff
@@ -21,7 +21,7 @@ trustedProfileAnalyzerRealm:
   clients:
     # The main TPA actor, responsible to parse the information from SBOMs and
     # other source into its database.
-    cli:
+    walker:
```
lingyzhuang, line 24: @Roming22 Can you please set it to `cli` also? I guess it's not used for configuration here, but I think it's better to make it the same with file  `installer/charts/tssc-tpa/values.yaml`. BTW, test can pass with both `cli` and `walker`

otaviof (approved):
/lgtm

installer/charts/tssc-tpa/values.yaml
Roming22: Done

installer/charts/tssc-tpa-realm/values.yaml
Roming22: Done

installer/charts/values.yaml.tpl
Roming22: Done

installer/charts/tssc-tpa/values.yaml
otaviof: It should be left as `walker`, as well all other references, please.

otaviof (changes requested):
The `values.yaml` on `tssc-tpa-realm` also need to reflect the "walker" client name: https://github.com/redhat-appstudio/rhtap-cli/blob/d7b349aff8bc883eb1e5c1e75c1f00a81c426e26/installer/charts/tssc-tpa-realm/values.yaml#L21-L31

installer/charts/tssc-tpa/values.yaml
```diff
@@ -97,12 +97,12 @@ redhat-trusted-profile-analyzer:
       # A specialized crawler to interact with external data sources, this client
       # is used by other TPA modules.
       cli:
```
otaviof, line 99: Here we also need this named `walker`, for the record.

otaviof (changes requested):

installer/charts/tssc-tpa/values.yaml
Roming22: Not for TPA v2.

Roming22:
After more discussions, this PR has been deemed as unnecessary.

PR #836, opened 2025-06-04
Do not enable ODF if TPA is enabled

lcarva:
TPA is configured to use filesystem storage. It does not have an ODF dependency.


dperaza4dustbit (approved):
/lgtm

PR #856, opened 2025-06-10
Remove pre-defined role for service accounts

lingyzhuang:
In `_common/_service-account.tpl` file:
1.  Create `ServiceAccount`
2. Create `ClusterRole` with permissions **read/create/delete/patch** to secrets. Bind this role to namespaces `.Release.Namespace` and `tssc`
3. Create **read-only** ClusterRole, do ClusterRoleBinding for this role
4. `RoleBinding` binds a role to a specific namespace (`.Release.Namespace`）

Remove `Cluster-admin` ClusterRole, use read-only ClusterRole and secret read-write ClusterRole instead. 

ServiceAccount in chart `tssc-gitops` needs all verbs to all the resources to do some configurations, so create  a `Role` and bind it to namespace tssc-gitops.

Jira: [RHTAP-4880](https://issues.redhat.com/browse/RHTAP-4880)


installer/charts/tssc-acs-test/templates/service-account.yaml
```diff
@@ -1,31 +1,9 @@
...
-    verbs: 
-      - get
-      - list
-      - create
-      - delete
-      - update
-      - patch
-
+{{- $namespaces := dict
+  "integrationsecretnamespace" .Values.acsTest.integrationSecret.namespace
```
Roming22: I would rename `integrationSecret` to `tssc` in the yaml. It also needs to be added to the `config.yaml` of the chart itself.
I would rename the Helm var accordingly.

I think this could be simplified if you changed the key in the `config.yaml` to match the chart name (eg `acs` -> `tssc-acs`). You could then use something like `dig .Chart.Name "tssc" "namespace" .` to get the namespace directly in the `_common/` file.

lingyzhuang: @Roming22 by `config.yaml` do you mean `values.yaml`? If change the key in `values.yaml`, that will lead to many other changes, e.g. if I change `argoCD` to `tssc-gitops`, following updates are also needed:
```./templates/NOTES.txt:{{- if .Values.argoCD.enabled }}
./templates/NOTES.txt:  {{- $argoCD := .Values.argoCD -}}
./templates/_argocd.tpl:  {{- $argoCD := .Values.argoCD -}}
./templates/_argocd.tpl:  {{ printf "%s-cluster" .Values.argoCD.name }}
./templates/_argocd.tpl:  {{- $argoCD := .Values.argoCD -}}
./templates/argocd.yaml:{{- if .Values.argoCD.enabled }}
./templates/argocd.yaml:  {{- $argoCD := .Values.argoCD -}}
./templates/job-post-deploy.yaml:{{- if .Values.argoCD.enabled }}
./templates/job-post-deploy.yaml:  {{- $argoCD := .Values.argoCD }}
./templates/tests/test.yaml:{{- if .Values.argoCD.enabled }}
./templates/tests/test.yaml:  {{- $argoCD := .Values.argoCD }}```

Do you think it's better to change the key in `values.yaml`?

Comment (Roming22) [reply installer/charts/tssc-acs-test/templates/service-account.yaml]: Correct, that is what I meant. You are right, that is a lot of changes. How about you just rename `integrationsecretnamespace` to `tssc` for the moment, and you do the renaming/simplification in a follow-up PR?

File: installer/charts/values.yaml.tpl [hunk 2]
```diff
@@ -205,6 +205,8 @@ argoCD:
 
 pipelines:
   namespace: {{ $pipelines.Namespace }}
+  integrationSecret:
```
Roming22: Please rename `integrationSecret` to `tssc` everywhere and we'll be good to go.

lingyzhuang: Done.

Roming22 (approved):
/lgtm

integration-tests/scripts/install.sh
```diff
@@ -53,18 +54,17 @@ github_integration() {
...
-    id: "${GITHUB__APP__ID}"
-    clientId: "${GITHUB__APP__CLIENT__ID}"
-    clientSecret: "${GITHUB__APP__CLIENT__SECRET}"
-    host: "github.com"
-    publicKey: |-
-$(printf "%s\n" "${GITHUB__APP__PRIVATE_KEY}" | sed 's/^/      /')
-    token: "${GITOPS__GIT_TOKEN}"
-    webhookSecret: "${GITHUB__APP__WEBHOOK__SECRET}"
-EOF
+    sed -i "/integrations:/ a \  github:\n\
```
Roming22, line 69: I feel like this would be much more readable if we used a tool like `yq` instead of `sed`.

Roming22 (approved):
/lgtm

PR #857, opened 2025-06-10
feat: update TektonChain configuration

Roming22:
Use the operator configuration to generate the signing-secret. Tests have shown that the operator does not overwrites an existing secret, so no additional safeguards are needed to enable the feature.

Remove the option to configure the secret name, as the name is static for TektonChain.

rh-pre-commit.version: 2.3.2

### DIFF
--- a/installer/charts/tssc-pipelines/scripts/test-signing-secrets.sh
+++ b/installer/charts/tssc-pipelines/scripts/test-signing-secrets.sh
@@ -1,0 +1,1 @@
+../../../scripts/test-signing-secrets.sh
--- a/installer/charts/tssc-pipelines/templates/_copy-scripts.tpl
+++ b/installer/charts/tssc-pipelines/templates/_copy-scripts.tpl
@@ -1,0 +1,1 @@
+../../_common/_copy-scripts.tpl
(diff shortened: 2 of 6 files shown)


installer/charts/tssc-infrastructure/templates/openshift-pipelines/tektonconfig/patch._tpl
```diff
@@ -1,6 +1,7 @@
 {{- define "pipelines.TektonConfigPatch" -}}
 spec:
   chain:
+    generateSigningSecret: true
```
lingyzhuang: Should it be `'true'`? The same as line 5.

Roming22: It believe it can take both, I've updated for consistency.

lingyzhuang (approved):

installer/charts/tssc-infrastructure/templates/openshift-pipelines/tektonconfig/patch._tpl
lingyzhuang: Sorry about the comment. I tested with this PR and the value should be type **bool**, not string. So please remove `''`.

lingyzhuang (changes requested):
Please change it back to `true`, instead of `'true'`.

installer/charts/tssc-infrastructure/templates/openshift-pipelines/tektonconfig/patch._tpl
Roming22: Done

lingyzhuang (approved):

PR #858, opened 2025-06-10
fix: default TektonConfig causing trouble

Roming22:
Despite the subscription configuration, the default TektonConfig is instanciated, causing the deployment to fail.

So we change the configuration so that it is created, and then patched during the deployment.

cf [RHTAP-5228](https://issues.redhat.com//browse/RHTAP-5228)


installer/charts/tssc-pipelines/templates/tektonconfig/patch_tekton_config.yaml
```diff
@@ -0,0 +1,43 @@
...
+            - |
+              if [[ "{{ .Values.subscriptions.openshiftPipelines.managed }}" != "true" ]]; then
+                  echo "Skipping TektonConfig resource as pipeline subscription is not managed"
+                  echo "Success"
+                  exit 0
+              fi
+
+              if ! oc get tektonconfig config >/dev/null; then
+                echo "Waiting for the default TektonConfig to be instanciated"
+                sleep 10
```
lcarva, line 37: Is the sleep right before `exit 1` meant to work together with `restartPolicy` to emulate a loop?

lcarva (approved):

PR #1027, opened 2025-07-16
RHTAP-5269: Mark Required Flags

otaviof:
Using Cobra to mark persistent flags, allowing a pragmatic approach to identify the required flags per subcommand.

## Summary by CodeRabbit

* **New Features**
  * Required flags for integrations are now explicitly enforced, ensuring users provide necessary information when configuring integrations.
  * Some integrations now have improved default values for certain options, enhancing the initial setup experience.

* **Refactor**
  * Improved flag registration across all integrations for greater consistency and reliability when using command-line options.

* **Style**
  * Minor formatting improvements for code readability.


Roming22:
@otaviof I'm not familiar with the library, but my understanding is that the code you've added would throw an error if the argument is missing.

If that is correct, then you need to clean up the `Validate` methods, as they're currently taking care of this check.

otaviof:
> @otaviof I'm not familiar with the library, but my understanding is that the code you've added would throw an error if the argument is missing.
> 
> If that is correct, then you need to clean up the `Validate` methods, as they're currently taking care of this check.

Yes, these methods emit errors, and it is good practice to catch them. However, the validation is still applicable; with this PR, we only make it possible to discover flags marked as required pragmatically. So, no need to clean up, IMO.

Roming22:
@otaviof I'm not sure I understand... Here's the current behavior with your change:
```
10:31:40 - RARNAUD:~/Code/rhads/cli [cobra-required]
> bin/tssc integration quay --token "REDACTED"
Error: dockerconfigjson is required
```

If I remove the validation in QuayIntegration.Validate(), we still get a meaningful error message:
```
10:34:19 - RARNAUD:~/Code/rhads/cli [cobra-required] (S:0 M:1 U:0)
> bin/tssc integration quay --token "REDACTED"
Error: required flag(s) "dockerconfigjson", "url" not set
```

So why not clean-up the Validate methods when they only check if the arg has been set?

Roming22:
@otaviof The PR also breaks args with default values:
```
> bin/tssc integration quay --dockerconfigjson 'foobar' --token "REDACTED"
Error: required flag(s) "url" not set
```

Roming22 (changes requested):
See my above comments.

otaviof:
> @otaviof I'm not sure I understand... Here's the current behavior with your change:
> 
> ```
> 10:31:40 - RARNAUD:~/Code/rhads/cli [cobra-required]
> > bin/tssc integration quay --token "REDACTED"
> Error: dockerconfigjson is required
> ```
> 
> If I remove the validation in QuayIntegration.Validate(), we still get a meaningful error message:
> 
> ```
> 10:34:19 - RARNAUD:~/Code/rhads/cli [cobra-required] (S:0 M:1 U:0)
> > bin/tssc integration quay --token "REDACTED"
> Error: required flag(s) "dockerconfigjson", "url" not set
> ```
> 
> So why not clean-up the Validate methods when they only check if the arg has been set?

Now I see. I'm removing the redundant validation. Thanks for pointing this out, the erro message created by Cobra is better than we had in place.

PR #1044, opened 2025-07-22
Fix the acs-quay assertion

lcarva:
Since Quay is no longer installed by the installer, there isn't an entry for it under integrations. Instead, the presence of the quay secret in the cluster is sufficient.

## Summary by CodeRabbit

* **Refactor**
  * Updated the conditions for creating the Quay integration container to require ACS integration to be enabled and the Quay secret to exist.

### DIFF
--- a/installer/charts/tssc-integrations/templates/acs/pod.yaml
+++ b/installer/charts/tssc-integrations/templates/acs/pod.yaml
@@ -90,7 +90,7 @@ spec:
         allowPrivilegeEscalation: false
   {{- end }}
   {{- $secretQuay := (lookup "v1" "Secret" .Release.Namespace "tssc-quay-integration") }}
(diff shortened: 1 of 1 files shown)


dperaza4dustbit (approved):
/lgtm

lcarva:
For posterity: the issue fixed by this PR only affects those that didn't have ACS enabled. If you did have ACS enabled, the second term of the `or` condition was never evaluated, thus not causing a problem.

The error I was seeing (with ACS disabled) was:
```
############################################################
# [12/12] Deploying 'charts/tssc-integrations' in 'tssc'.
############################################################
Error: install failed: template: tssc-integrations/templates/acs/pod.yaml:93:56: executing "tssc-integrations/templates/acs/pod.yaml" at <$integrations.quay.enabled>: nil pointer evaluating interface {}.enabled
```

Roming22 (approved):
/lgtm

PR #1086, opened 2025-07-30
Adding coderabbit config file

dperaza4dustbit:
Also changing the path_filters list to include a vendors exclude filter.

## Summary by CodeRabbit

* **Chores**
  * Added a new configuration file to centralize settings for automated code review, linting, chat integration, knowledge management, and code generation workflows.
  * Configured the review process to exclude files in vendor directories.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -1,0 +1,4 @@
+---
+reviews:
+  path_filters:
+    - "!vendor/**"


.coderabbit.yaml
Roming22, line 1: Can we just override the defaults? I would like to avoid having to sync this file from time to time.

dperaza4dustbit: I read this as there is a set of defaults but no necessarily what our repo has now: https://docs.coderabbit.ai/guides/repository-settings#about
Also when I follow these instructions the tips recommend copying the config from the UI: https://docs.coderabbit.ai/getting-started/configure-coderabbit#configure-coderabbit-using-a-yaml-file

Roming22: I don't see anything that says that all the values are needed. Could you save the current config, submit a PR with just the filtering, then check if the config returned by CodeRabbit is what we expect? If it isn't then this PR can be merged.

dperaza4dustbit: Will test that OK

dperaza4dustbit:
@coderabbitai configuration

dperaza4dustbit:
@coderabbitai configuration

.coderabbit.yaml
dperaza4dustbit: OK adding only the path_filter section on config file results in same configuration reported as having all config options. I think we should be fine with just that review.path_filter config @Roming22

Roming22 (approved):
/lgtm

PR #1135, opened 2025-08-13
Remove subscription to Crunchy Postgres

dperaza4dustbit:
Implements [RHTAP-5355](https://issues.redhat.com//browse/RHTAP-5355)

Since there is no instance for cruncy pg at the moment there is no need to install the operator.

## Summary by CodeRabbit

- Chores
  - Removed the Crunchy Data PostgreSQL Operator from default subscriptions; it will no longer be installed or managed by the installer.
  - Removed the RBAC resources that were previously created for Crunchy Data.
  - Other subscriptions (GitOps, Keycloak, Pipelines, Trusted Artifact Signer, Advanced Cluster Security, Developer Hub) remain unchanged.


Roming22 (approved):
/lgtm

PR #1138, opened 2025-08-14
Take care of warning while deploying integrations

dperaza4dustbit:
Since the resources is not being used, removing to avoid noice with the warning. The requests section should be under resources

## Summary by CodeRabbit

* **Chores**
  * Reduced installation resource footprint by removing resource requests from a no-op container in the ACS post-install process. This lowers reserved CPU, memory, and ephemeral storage during install/upgrade.
  * Eases scheduling on resource-constrained clusters by avoiding unnecessary reservation for the post-install step.
  * No functional changes to ACS behavior; only installation overhead is reduced.

### DIFF
--- a/installer/charts/tssc-integrations/templates/acs/pod.yaml
+++ b/installer/charts/tssc-integrations/templates/acs/pod.yaml
@@ -129,10 +129,6 @@ spec:
         - bash
         - -c
         - "echo 'No op: Success'"
-      requests:
-        cpu: 125m
-        memory: 128Mi
-        ephemeral-storage: "100Mi"
       securityContext:
         allowPrivilegeEscalation: false
   {{- end }}


installer/charts/tssc-integrations/templates/acs/pod.yaml
```diff
@@ -129,10 +129,6 @@ spec:
         - bash
         - -c
         - "echo 'No op: Success'"
-      requests:
```
Roming22, line 132: I think it should be moved to the right place instead of being deleted, otherwise I believe it makes Snyk or SonarQube unhappy.

Roming22: SonarQube is happy, I don't see Snyk at the moment, so let's ignore my comment.

Roming22 (approved):

PR #1154, opened 2025-08-18
RHTAP-4050: Refactoring and Single Dependency Deploy Fix

otaviof:
Moving `config.Dependency` to the `resolver` package, and fixing single dependency (chart) deployment on `tssc deploy`.

## Summary by CodeRabbit

- Refactor
  - Switched to a unified dependency model across resolver, installer, hooks, topology and subcommands; topology APIs and dependency accessors updated.
  - Removed legacy exported dependency types and fields.
- Bug Fixes
  - Deploy now fails early with a clear error if chart files cannot be loaded.
- Documentation
  - Topology header/description updated from “Chart” to “Dependency”.
- UX
  - Deployment logs now include chart-path for better context.
- Tests
  - Added/updated tests to validate the new dependency model and APIs.


pkg/resolver/resolver.go
```diff
@@ -146,24 +150,23 @@ func (r *Resolver) resolveDependencies() error {
 			}
 			requiredBy = dependsOn
```
dperaza4dustbit, line 151: This is a bit confusing, if chart A dependsOn Chart B then it does not make sense to say that chart A is requiredBy chart B. On the other had, it does make sense to say chart B is requiredBy chart A. Not saying the logic is wrong but just the wording. Maybe it would be better to say if chart A dependsOn chart B then chart A requires (not required by) chart B. Also be careful with this assumption cause if chart B could be an integration instead of a chart then chart A could still be deployed without chart B.

otaviof: Well, I think the "required" word is synonymous to "depends on", a chart is required due to being a dependency.

pkg/installer/installer.go
```diff
@@ -14,15 +14,18 @@ import (
 	"github.com/redhat-appstudio/tssc-cli/pkg/k8s"
 	"github.com/redhat-appstudio/tssc-cli/pkg/monitor"
 	"github.com/redhat-appstudio/tssc-cli/pkg/printer"
+	"github.com/redhat-appstudio/tssc-cli/pkg/resolver"
 
 	"helm.sh/helm/v3/pkg/chartutil"
 )
 
+// Installer represents the "helm install" using its APIs, this component deploys
+// the informed dependency on the pre-configure namespace.
```
Roming22: pre-configure -> pre-configured

pkg/resolver/dependency.go
```diff
@@ -0,0 +1,99 @@
...
+func (d *Dependency) Namespace() string {
+	return d.namespace
+}
+
+// SetNamespace sets the namespace for this dependency.
+func (d *Dependency) SetNamespace(namespace string) {
+	d.namespace = namespace
+}
+
+// DependsOn returns a slice of dependencies names from  chart's annotation.
```
Roming22: from  chart's -> from  the chart's

pkg/resolver/resolver.go
```diff
@@ -24,58 +22,62 @@ var ErrCircularDependency = fmt.Errorf("circular dependency detected")
...
 
-// dependency returns the dependency for a chart. If the chart is associated with
-// a product, the dependency use the product namespace. If the chart contains a
-// "use-product-namespace" annotation, the dependency use the product namespace.
-// If no product is associated with the chart, the dependency use the installer
-// namespace.
-func (r *Resolver) dependency(hc *chart.Chart) (*config.Dependency, error) {
+// setDependencyNamespace sets the desired namespace on the informed dependency.
+// Regular charts are deployed on the same namespace than the installer, while
+// product related uses the what's configured for it.
```
Roming22: this part of the sentence is not correct.

pkg/resolver/resolver.go
```diff
@@ -24,58 +22,62 @@ var ErrCircularDependency = fmt.Errorf("circular dependency detected")
...
+func (r *Resolver) setDependencyNamespace(d *Dependency) error {
 	var product string
 	// Check if the Helm chart should use the product namespace.
-	if p := r.collection.UseProductNamespace(hc); p != "" {
+	if p := d.UseProductNamespace(); p != "" {
 		product = p
 	}
 	// Check if the Helm chart is associated with a product, which takes
 	// precedence the "use-product-namespace" annotation.
```
Roming22: precedence the -> precedence over the

pkg/resolver/resolver.go
```diff
@@ -102,14 +104,16 @@ func (r *Resolver) dependsOn(
 // resolveEnabledProducts resolves the dependencies of enabled products.
 func (r *Resolver) resolveEnabledProducts() error {
 	for _, product := range r.cfg.GetEnabledProducts() {
-		hc, err := r.collection.GetProductChart(product.Name)
+		d, err := r.collection.GetProductDependency(product.Name)
 		if err != nil {
 			return err
 		}
+		// Products uses the namespace specified in configuration.
```
Roming22: in configuration -> in the configuration

pkg/resolver/resolver.go
```diff
@@ -120,22 +124,22 @@ func (r *Resolver) resolveEnabledProducts() error {
...
-		for _, dependsOn := range r.collection.DependsOn(&hc) {
-			// Ensure the required chart is in the topology, when not in the
+		for _, dependsOn := range d.DependsOn() {
+			// Ensure the required dependency is in the topology, when not in the
 			// topology it is skipped.
 			if !r.topology.Contains(dependsOn) {
 				continue
 			}
-			// Ensures the if the required chart is in the collection.
+			// Ensures the if the required dependency is in the collection.
```
Roming22: Ensures the if?

pkg/resolver/resolver.go
dperaza4dustbit: Don't think so, (A depends on B) != (A required by B) however (A depends on B) == (A requires B)

A required by B means B needs A.
A requires B means A needs B

It is an English language nuance but I think I'm right.

Roming22: I'm with David. If A needs B, then A `requires` B and B is `required by` A. In other words:
- `requires` = `depends on`
- `required` = `needed by`

pkg/installer/installer.go
otaviof: Done, thanks!

Roming22: Still there

pkg/resolver/resolver.go
```diff
@@ -24,58 +22,62 @@ var ErrCircularDependency = fmt.Errorf("circular dependency detected")
...
 
-// dependency returns the dependency for a chart. If the chart is associated with
-// a product, the dependency use the product namespace. If the chart contains a
-// "use-product-namespace" annotation, the dependency use the product namespace.
-// If no product is associated with the chart, the dependency use the installer
-// namespace.
-func (r *Resolver) dependency(hc *chart.Chart) (*config.Dependency, error) {
+// setDependencyNamespace sets the desired namespace on the informed dependency.
+// By default, charts are deployed on the same namespace than the installer, while
+// product assossiated dependencies will use the namespace configured for the it.
```
Roming22: configured for the it?

Roming22 (commented):
Still a couple of typos left out.

Roming22 (approved):
/lgtm

PR #1157, opened 2025-08-19
Renaming the required chart variable

dperaza4dustbit:
RequiredBy variable implies that current chart is required by the dependency but it is the other way around, changing to avoid confusion in the future. Logic does not need to change as they are correct.

## Summary by CodeRabbit

* **Refactor**
  * Internal code readability improved through non-functional renaming and cleanup.
  * Dependency resolution behavior, ordering, and public interfaces remain unchanged.
  * Backward compatibility preserved; no user-facing changes.


Roming22 (approved):
/lgtm

PR #1190, opened 2025-08-27
fix: Add unittests to pipelines

Roming22:
## Summary by CodeRabbit

- **Tests**
  - Updated test expectations to reflect revised dependency resolution ordering and namespace mappings.
  - Ensured all tests run as part of the image build process.

- **Chores**
  - Included test assets in the Docker build context.
  - Added a build step to execute the test suite before compiling, improving build-time validation.


pkg/resolver/resolver_test.go
```diff
@@ -50,29 +50,29 @@ func TestNewResolver(t *testing.T) {
...
+			"tssc-tpa":            "tssc-tpa",
+			"tssc-tas":            "tssc-tas",
+			"tssc-pipelines":      "tssc",
+			"tssc-gitops":         "tssc-gitops",
+			"tssc-app-namespaces": "tssc",
+			"tssc-dh":             "tssc-dh",
+			"tssc-acs":            "tssc-acs",
+			"tssc-acs-test":       "tssc-acs",
+			"tssc-integrations":   "tssc",
 		}))
```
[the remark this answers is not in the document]
Roming22: That's the reference to the deployment namespace, but the change is related to the namespace where the helm chart is registered.

pkg/resolver/resolver_test.go
```diff
@@ -50,29 +50,29 @@ func TestNewResolver(t *testing.T) {
 		// Validating the order of the resolved dependencies, as well as the
 		// namespace of each dependency.
 		g.Expect(dependencyNamespaceMap).To(o.Equal(map[string]string{
-			"tssc-openshift":        "tssc",
-			"tssc-subscriptions":    "tssc",
-			"tssc-infrastructure":   "tssc",
-			"tssc-iam":              "tssc-keycloak",
```
dperaza4dustbit, line 56: Why did you change the namespace?

Dockerfile
```diff
@@ -12,10 +12,12 @@ COPY installer/ ./installer/
 
 COPY cmd/ ./cmd/
 COPY pkg/ ./pkg/
+COPY test/ ./test/
 COPY vendor/ ./vendor/
 
 COPY go.mod go.sum Makefile ./
 
+RUN make test
```
otaviof, line 20: Unit testing should run outside of the container image creation. CI should have a job to exercise it.

otaviof (changes requested):
Could we move it to a CI job?

pkg/resolver/resolver_test.go
Roming22: Because the test fails without that change. My understanding is that the namespace that is being returned is not the namespace where the chart is deployed, but the namespace where the chart is registered.

Dockerfile
Roming22: I've looked into it because I thought it would be simple. It wasn't and this change is and achieves what needs to be done. While I agree it would be prettier if it was its own thing, I believe it is good enough.

Do you see any technical drawbacks with this implementation? My feeling is this test is cheap to run so the impact on the build time is negligible.

pkg/resolver/resolver_test.go
dperaza4dustbit: I see so the iam chart release is going to tssc while manifests are going to tssc-keycloak. OK

dperaza4dustbit (approved):
/lgtm

Dockerfile
Roming22: Here's a few reasons why I went this way:

1. Doesn't require us to change the default Konflux pipeline, which means we can benefit from auto-upgrades to the pipeline.
2. If it was separate, we'd need to keep the build image and test image in sync. That seems like a headache for little added value.
3. It took more than 15m to figure out how to access the code in the Konflux pipeline, and after thinking about the above pain points, I decided it wasn't worth it.

otaviof: > Do you see any technical drawbacks with this implementation? My feeling is this test is cheap to run so the impact on the build time is negligible.

Unit tests must stop the container image build. This means a basic application component isn't working and must stop the pipeline as soon as possible. Before the container image build, and specially, before end-to-end testing.

As you noticed, the unit-testing is validating the expected topology, which means is fundamental to run unit testing as close as possible to the application executable build.

As you mentioned, they are rather fast to run and, given their importance, must be the first.

lcarva: > Doesn't require us to change the default Konflux pipeline, which means we can benefit from auto-upgrades to the pipeline.

It's a bit more complex than this. The Pipeline definition is duplicated in this repo in two places under the `.tekton` directory. When a new Task is added to the default Konflux Pipeline, it doesn't automatically make its way here. This is definitely a pain point when using Konflux. The individual Task references do get updated with mintmaker via PRs, but nothing will automatically submit a PR to add a new Task. (Unless there's a direct effort to do this which has happened in the past but it's not standard.)


> If it was separate, we'd need to keep the build image and test image in sync. That seems like a headache for little added value.

I believe the [run-script](https://github.com/konflux-ci/build-definitions/tree/main/task/run-script-oci-ta/0.1) Task is something that could be added to the pipeline to run Make, for example. That does require as a parameter an image to run the command on. Looking at the existing Dockerfile, I think `registry.access.redhat.com/ubi9/go-toolset:1.24.6-1755755147` should be usable as is. So no need for us to maintain a new image.

> It took more than 15m to figure out how to access the code in the Konflux pipeline, and after thinking about the above pain points, I decided it wasn't worth it.

I'll admit it's not the easiest. But starting by looking at the `.tekton` directory is always the best starting point.

With all that, I don't think it's terribly wrong to run `make test`  in the Dockerfile because this is happening in a multi-stage build. I went searching for what other folks are saying and this seems to be one of those things that folks don't all agree on. For example, here's a [guide](https://docs.docker.com/guides/java/run-tests/) directly from docker on how to run tests in a multi-stage build.

I see the pros and cons of either approach. If my input matters, I'd suggest using whatever makes sense right now. These things are not set in stone and we can evolve them when it's obvious the current approach doesn't work for us anymore.

PR #1233, opened 2025-09-09
Umarshal config.yaml to yaml.Node

lingyzhuang:
Remove `payload` in type config.Config add `root yaml.Node instead,` . Unmarshal config into *yaml.Node, decode it to Installer, marshal it to payload, then update the configuration in cluster.

Jira: [RHTAP-5316](https://issues.redhat.com/browse/RHTAP-5316)

## Summary by CodeRabbit

- New Features
  - Improved YAML config handling: load from file/bytes and decode in-memory YAML for reliable processing and round-trips.

- Bug Fixes
  - Better validation and clearer errors for empty/invalid configs.
  - Dry-run no longer prints configuration when an error occurs.

- Refactor
  - Encoding/decoding path revamped to preserve YAML structure and emit consistent document headers.

- Tests
  - Added/updated tests for decode/encode round-trips and string output equality.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 
@@ -29,10 +30,9 @@ type Spec struct {
 
 // Config root configuration structure.
 type Config struct {
-	cfs     *chartfs.ChartFS // embedded filesystem
-	payload []byte           // original configuration payload
-
-	Installer Spec `yaml:"tssc"` // root configuration for the installer
+	cfs       *chartfs.ChartFS // embedded filesystem
+	Installer Spec             `yaml:"tssc"` // root configuration for the installer
+	root      yaml.Node        // yaml data representation
 }
 
 var (
@@ -90,18 +90,47 @@ func (c *Config) Validate() error {
 	return nil
 }
 
+// DecodeNode returns a struct converted from *yaml.Node
+func (c *Config) DecodeNode() error {
+	if len(c.root.Content) == 0 {
+		return fmt.Errorf("invalid configuration: content is empty")
+	}
+	if len(c.root.Content[0].Content) == 0 {
+		return fmt.Errorf("invalid configuration: missing content")
+	}
+	root := c.root.Content[0].Content[1]
+	if err := root.Decode(&c.Installer); err != nil {
+		return err
+	}
+	return nil
+}
+
 // MarshalYAML marshals the Config into a YAML byte array.
 func (c *Config) MarshalYAML() ([]byte, error) {
-	return yaml.Marshal(c)
+	var buf bytes.Buffer
+	if len(c.root.Content) == 0 {
+		return nil, fmt.Errorf("invalid configuration format: content is nil or empty")
+	}
+	buf.WriteString("---\n")
+	encoder := yaml.NewEncoder(&buf)
+	encoder.SetIndent(2)
+	defer encoder.Close()
+	if err := encoder.Encode(c.root.Content[0]); err != nil {
+		return nil, err
+	}
+	return buf.Bytes(), nil
 }
 
 // UnmarshalYAML Un-marshals the YAML payload into the Config struct, checking the
 // validity of the configuration.
-func (c *Config) UnmarshalYAML() error {
-	if len(c.payload) == 0 {
+func (c *Config) UnmarshalYAML(payload []byte) error {
+	if len(payload) == 0 {
 		return ErrEmptyConfig
 	}
-	if err := yaml.Unmarshal(c.payload, c); err != nil {
+	if err := yaml.Unmarshal(payload, &c.root); err != nil {
+		return fmt.Errorf("%w: %w", ErrUnmarshalConfig, err)
+	}
+	if err := c.DecodeNode(); err != nil {
 		return fmt.Errorf("%w: %w", ErrUnmarshalConfig, err)
 	}
 	return c.Validate()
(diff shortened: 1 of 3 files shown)


pkg/config/config.go
```diff
@@ -35,6 +36,10 @@ type Config struct {
 	Installer Spec `yaml:"tssc"` // root configuration for the installer
 }
 
+type Node struct {
+	root *yaml.Node
```
otaviof: The `yaml.Node` instance must live in the `Config` itself. That's where we store the configuration payload.

pkg/config/config.go
```diff
@@ -97,30 +102,162 @@ func (c *Config) MarshalYAML() ([]byte, error) {
...
 	}
-	if err := yaml.Unmarshal(c.payload, c); err != nil {
-		return fmt.Errorf("%w: %w", ErrUnmarshalConfig, err)
+	var node yaml.Node
+	if err := yaml.Unmarshal(c.payload, &node); err != nil {
+		return nil, fmt.Errorf("%w: %w", ErrUnmarshalConfig, err)
 	}
-	return c.Validate()
+	return &node, nil
+}
```
otaviof: Here, following my previous comment, the `UnamarshalYAML` should write to the `*Config`, itself.

pkg/config/config.go
```diff
@@ -97,30 +102,162 @@ func (c *Config) MarshalYAML() ([]byte, error) {
...
+		}
+		keyPath, value := parts[0], parts[1]
+		keyPathValues[keyPath] = value
+	}
+	if len(keyPathValues) == 0 && len(setValues) > 0 {
+		return nil, fmt.Errorf("No valid updates applied")
+	}
+	return keyPathValues, nil
+}
+
```
otaviof: Also, it must write, or update, the information on the `*Config` object.

pkg/config/config.go
```diff
@@ -129,8 +266,12 @@ func NewConfigFromFile(cfs *chartfs.ChartFS, configPath string) (*Config, error)
...
 func NewConfigFromBytes(payload []byte) (*Config, error) {
 	c := &Config{payload: payload}
-	if err := yaml.Unmarshal(payload, c); err != nil {
-		return nil, fmt.Errorf("%w: %s", ErrUnmarshalConfig, err)
+	node, err := c.UnmarshalYAML()
+	if err != nil {
+		return nil, err
+	}
+	if err = c.DecodeNode(node); err != nil {
+		return nil, err
```
otaviof, line 176: Same here.

pkg/config/config.go
```diff
@@ -144,3 +285,24 @@ func NewConfigDefault() (*Config, error) {
...
+	}
+	if err = confNode.UpdateInConfig(keyPathValues); err != nil {
+		return nil, err
+	}
+	if err = c.DecodeNode(confNode.root); err != nil {
+		return nil, err
+	}
+	c.payload, err = yaml.Marshal(confNode.root)
+	return c, nil
+}
```
otaviof: The `*Config` object should be able to receive updates, same than the previous comments.

pkg/config/config_test.go
otaviof, line 1: I'd like to see more testing for the update feature, please.

pkg/subcmd/config.go
otaviof, line 1: Let's leave the subcommand changes for another PR, please. I'd like to work with you on this change.

otaviof (changes requested):
Let's focus on the `config.Config` struct to support `yaml.Node` instead of the `payload []byte` it's currently using. Let's leave the `tssc config` subcommand to a follow up PR, please.

pkg/config/config.go
```diff
@@ -90,46 +89,70 @@ func (c *Config) Validate() error {
...
+func (c *Config) UnmarshalYAML(payload []byte) error {
+	if len(payload) == 0 {
 		return ErrEmptyConfig
 	}
-	if err := yaml.Unmarshal(c.payload, c); err != nil {
+	if err := yaml.Unmarshal(payload, &c.root); err != nil {
 		return fmt.Errorf("%w: %w", ErrUnmarshalConfig, err)
 	}
-	return c.Validate()
+	return nil
```
otaviof: Here I think you'll need to "unmarshal" into the `config.Installer` struct, and you should continue to invoke the `c.Validate()`.

pkg/config/config.go
```diff
@@ -90,46 +89,70 @@ func (c *Config) Validate() error {
...
 }
 
 // String returns this configuration as string, indented with two spaces.
-func (c *Config) String() string {
-	return string(c.payload)
+func (c *Config) String(payload []byte) string {
+	return string(payload)
```
otaviof: The `String()` method must return the self representation nas string, instead of receiving a parameter.

otaviof (changes requested):

pkg/config/config.go
```diff
@@ -29,10 +30,9 @@ type Spec struct {
 
 // Config root configuration structure.
 type Config struct {
-	cfs     *chartfs.ChartFS // embedded filesystem
-	payload []byte           // original configuration payload
-
-	Installer Spec `yaml:"tssc"` // root configuration for the installer
+	cfs       *chartfs.ChartFS // embedded filesystem
+	Installer Spec             `yaml:"tssc"` // root configuration for the installer
+	root      yaml.Node        // unmarshal yaml config to yaml.Node
```
otaviof: nit: the comment about this item should be "yaml data representation".

pkg/config/config.go
```diff
@@ -90,46 +90,79 @@ func (c *Config) Validate() error {
...
+	if len(c.root.Content[0].Content) == 0 {
+		return fmt.Errorf("invalid configuration: missing content")
+	}
+	root := c.root.Content[0].Content[1]
+	if err := root.Decode(&c.Installer); err != nil {
+		return err
+	}
+	return nil
+}
+
```
otaviof, line 117: Where's this function being used?

otaviof (approved):
/lgtm

PR #1234, opened 2025-09-09
RHTAP-4050: Refactoring Integrations

otaviof:
Refactoring the integrations to rely on a single actor to manage Kubernetes secrets, and split the business logic responsibility on dedicated components. This pull-request introduces a central integrations manager to be used later on for MCP tools.

## Summary by CodeRabbit

- New Features
  - Added unified integrations for ACS, Azure, Bitbucket, GitHub App, GitLab, Jenkins, and container registries (Artifactory, Nexus, Quay) with consistent CLI flags and validations.
  - Introduced an Integration Manager to register and manage integrations centrally.

- Refactor
  - Consolidated per-integration workflows into a shared integration provider; subcommands now use the common provider for secret creation and logging.

- Chores
  - Removed legacy per-integration implementations in favor of the unified model.


pkg/integration/containerregistry.go
Roming22, line 1: Semantic nit picking, but I believe this should be `imageregistry`, as a container is an instance of an image.
Note that the string "Container" appears a few time and would need to be replaced as well.

pkg/integration/githubapp.go
Roming22, line 1: Since we have just that single integration for GitHub, how about just renaming it `github`, so it matches the other git providers?

pkg/integration/githubapp.go
```diff
@@ -0,0 +1,261 @@
...
+		"slug":          []byte(appConfig.GetSlug()),
+		"updatedAt":     []byte(appConfig.UpdatedAt.String()),
+		"webhookSecret": []byte(appConfig.GetWebhookSecret()),
+		"token":         []byte(g.token),
+		"username":      []byte(username),
+	}, nil
+}
+
+// NewGithubApp instances a new GitHub App integration.
+func NewGithubApp(logger *slog.Logger, kube *k8s.Kube) *GitHubApp {
```
Roming22: `NewGitHubApp`?

pkg/integrations/manager.go
```diff
@@ -0,0 +1,72 @@
...
+	BitBucket      IntegrationName = "bitbucket"
+	GitHubApp      IntegrationName = "github-app"
+	GitLab         IntegrationName = "gitlab"
+	Jenkins        IntegrationName = "jenkins"
+	Nexus          IntegrationName = "nexus"
+	Quay           IntegrationName = "quay"
+	Trustification IntegrationName = "trustification"
+)
+
+const (
```
Roming22, line 34: I'm curious, why is this declared here instead of `containerregistry.go`?

Roming22 (commented):
Just a couple of nit picks as per usual, but it looks beautiful 🤩

pkg/integrations/manager.go
otaviof: The default image registries hostnames are informed here to be passed to the `integration.ImageRegistry`, a central location to hold the defaults.

pkg/integration/githubapp.go
otaviof: Thanks!

pkg/integration/githubapp.go
otaviof: On this PR I'm keeping the GitHub-App, as in `tssc integration github-app` consistent, I think would be a good idea to simply rename it to "GitHub" on a subsequent PR.

pkg/integration/containerregistry.go
otaviof: Sure, I agree, the name "Image Registry" makes more sense. Just renamed.

pkg/integration/githubapp.go
Roming22: Okay.

pkg/integration/imageregistry.go
```diff
@@ -0,0 +1,89 @@
...
+var _ Interface = &ImageRegistry{}
+
+const dockerConfigEx = `{ "auths": { "registry.tld": { "auth": "username" } } }`
+
+// PersistentFlags adds the persistent flags to the informed Cobra command.
+func (i *ImageRegistry) PersistentFlags(cmd *cobra.Command) {
+	p := cmd.PersistentFlags()
+
+	p.StringVar(&i.dockerConfig, "dockerconfigjson", i.dockerConfig,
+		fmt.Sprintf("JSON formatted registry credentails, e.g.: %q",
```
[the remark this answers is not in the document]
Roming22: @otaviof This is a valid comment.

pkg/integrations/manager.go
Roming22: I won't hold the PR for that, but I'm afraid I don't understand. Let's remember to discuss it on our 1-1.

pkg/integration/imageregistry.go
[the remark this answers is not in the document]
otaviof: Indeed, just fixed.

Roming22 (approved):
/lgtm

PR #1237, opened 2025-09-10
fix: disable microsoft auth in DH

Roming22:
The feature was not complete and is low priority, so it is disabled for 1.7.


dperaza4dustbit (approved):
/lgtm

PR #1245, opened 2025-09-16
RHTAP-4050: Integrations Resolver

otaviof:
## Summary by CodeRabbit

- New Features
  - Helm charts can declare provided and required integrations; deployment validates and enforces them.
  - Topology now reports Provided-Integrations and Required-Integrations; new topology builder powers resolution and command flows.
  - GitHub integration command renamed to "integration github" (alias preserved).

- Documentation
  - Docs updated to describe the two new integration annotations and examples.

- Tests
  - Unit tests added for integration expression parsing and helper parsing.

- UX/Style
  - Improved status/error messages for unresolved dependencies and integrations; disclaimer spacing adjusted.


docs/topology.md
```diff
@@ -47,6 +47,28 @@ annotations:
...
+
+### `tssc.redhat-appstudio.github.com/integrations-required`
+
+- **Purpose**: This **optional** annotation specifies the integrations that the Helm chart requires.  The value is a CEL expression that, when evaluated, determines if the required integrations are available.
+- **Usage**: The CEL expression uses the names of the integrations.  If the expression evaluates to `true`, the integrations are considered available.
+- **Example**:  If the chart requires both `git` and `registry` integrations:
+
+```yaml
+annotations:
+  tssc.redhat-appstudio.github.com/integrations-required: "github || gitlab"
```
Roming22: Example does not match the description on line 65.

installer/charts/tssc-app-namespaces/Chart.yaml
```diff
@@ -5,3 +5,4 @@ type: application
 version: "1.7.0"
 annotations:
   tssc.redhat-appstudio.github.com/depends-on: tssc-acs, tssc-pipelines, tssc-tpa
+  tssc.redhat-appstudio.github.com/integrations-required: "(github || gitlab) && acs && trustification"
```
Roming22: `(bitbucket || github || gitlab)`?

installer/charts/tssc-dh/Chart.yaml
```diff
@@ -7,3 +7,4 @@ appVersion: "1.7"
 annotations:
   tssc.redhat-appstudio.github.com/product-name: Developer Hub
   tssc.redhat-appstudio.github.com/depends-on: tssc-openshift, tssc-subscriptions, tssc-infrastructure, tssc-gitops, tssc-tas, tssc-pipelines, tssc-tpa, tssc-app-namespaces
+  tssc.redhat-appstudio.github.com/integrations-required: "(github || gitlab) && quay"
```
Roming22: (bitbucket || github || gitlab)?
(artifactory || nexus || quay)?

installer/charts/tssc-pipelines/Chart.yaml
```diff
@@ -8,3 +8,4 @@ appVersion: "1.19"
 annotations:
   tssc.redhat-appstudio.github.com/product-name: OpenShift Pipelines
   tssc.redhat-appstudio.github.com/depends-on: tssc-openshift, tssc-subscriptions, tssc-tas
+  tssc.redhat-appstudio.github.com/integrations-required: github
```
Roming22: I don't think that required. One should be able to install RHADS and use it with just GitLab or Bitbucket, which don't integrate in Pipelines in the same way as GitHub.

However, I think the `tas` integration is required, which you have forgotten to add to `tssc-tas`.

pkg/integrations/manager.go
```diff
@@ -1,52 +1,77 @@
...
-// IntegrationName name of a integration.
-type IntegrationName string
-
 const (
 	ACS            IntegrationName = "acs"
 	Artifactory    IntegrationName = "artifactory"
 	Azure          IntegrationName = "azure"
 	BitBucket      IntegrationName = "bitbucket"
-	GitHubApp      IntegrationName = "github-app"
+	GitHub         IntegrationName = "github"
```
Roming22, line 29: With this renaming, I believe some scripts need to be updated.

pkg/mcpserver/mcpserver.go
```diff
@@ -29,7 +29,7 @@ func (m *MCPServer) Start() error {
 func NewMCPServer() *MCPServer {
 	return &MCPServer{s: server.NewMCPServer(
 		constants.AppName,
-		"1.6.0",
+		"1.7.0",
```
Roming22: Please open a separate PR to udpate `hack/code-freeze.sh` so we don't forget to bump the version next time.

pkg/resolver/cel.go
```diff
@@ -0,0 +1,97 @@
...
+	// All expressions must evaluate to true, meaning all required integrations
+	// are configured.
+	if result.Value() == true {
+		return nil
+	}
+
+	// Using the referenced integration names to determine which integrations are
+	// missing, as in should be configured in the cluster but aren't found.
+	missing := []string{}
+	for _, ref := range referenced {
```
Roming22, line 73: I'm I right that if I have `(A || B) && (C || D)` and only A is configured, I'll get an error about B, C, D not being configured?
I can live with the limitation, but want to make sure I understood the code.

pkg/resolver/helper_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+func Test_commaSeparatedToSlice(t *testing.T) {
+	tests := []struct {
+		name           string
+		commaSeparated string
+		want           []string
+	}{{
+		name:           "empty string",
+		commaSeparated: "",
+		want:           nil,
```
Roming22, line 16: Why not an empty slice?

pkg/resolver/integrations.go
```diff
@@ -0,0 +1,105 @@
...
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/redhat-appstudio/tssc-cli/pkg/config"
+	"github.com/redhat-appstudio/tssc-cli/pkg/integrations"
+)
+
+// Integrations represents the actor the inspects the integrations provided and
```
Roming22: the actor that?

pkg/resolver/integrations.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	// When the integration exists, it marks the integration name as true, so it's
+	// configured in the cluster.
+	for _, name := range configuredIntegrations {
+		i.configured[name] = true
+	}
+	// Going through all valid integration names, by default when not registered
+	// the integration name is marked as false, as in not configured in the
+	// cluster.
+	for _, name := range manager.IntegrationNames() {
```
Roming22, line 95: That works and I'm not asking you to change it. I'm curious if you thought about instead of initializing the values of all the integrations, you'd only initialize the one that exists, and set the value to false for the missing ones in cel.Evaluate.

pkg/subcmd/integration_githubapp.go
Roming22, line 1: Shouldn't this file be renamed?

installer/charts/tssc-pipelines/Chart.yaml
otaviof: Cool, I'm removing the annotation from this chart. And `tas` is not a valid integration name, the ones supported are `tssc integration --help` names.

pkg/integrations/manager.go
otaviof: I've added `github-app` as a alias to `github`, thus you could use `tssc integration github` and `tssc integration github-app` interchangeably.

pkg/mcpserver/mcpserver.go
otaviof: Let me revert this change, I'm asking @raisavarghese to helps us to keep a central version number, that we can access from all components in the project. And on this story we have to have a look on the `code-freeze.sh` script too.

pkg/resolver/cel.go
otaviof: Yes, that's correct, the `missing` slice would receive `B`, `C` and `D` given these were options (variables) defined in the expression but not configured (false). Thus the user would have to consider the expression to understand which one, `B`, `C` or `D` would apply to the use-case.

pkg/resolver/helper_test.go
otaviof: In Golang nil is a empty slice, the internal language provides a data-structure that provides a view to a contiguous block of memory. So, when the slice descriptor is nil, the compiler consider that a empty slice in practice.

pkg/resolver/integrations.go
otaviof: Here we initialise the CEL program with all the valid options (variables), and in our case, these options are integration names set to false by default. Later on the component inspects the cluster, setting the configured integration options to true. So, we loop over all integration names initially to define the set of options a CEL program can reference.

pkg/subcmd/integration_githubapp.go
otaviof: Yes, that's correct!

docs/topology.md
otaviof: Thanks, it's now updated.

installer/charts/tssc-pipelines/Chart.yaml
Roming22: Ok, I won't let it delay this PR but we need a story to add that integration.

pkg/resolver/cel.go
Roming22: Good enough for me at the moment.

pkg/subcmd/integration_github.go
```diff
@@ -71,37 +70,35 @@ func (g *IntegrationGitHubApp) Complete(args []string) error {
...
-			Namespace: g.cfg.Installer.Namespace,
-			Name:      "tssc-github-integration",
-		})
+		return g.integration.Create(g.cmd.Context(), g.cfg)
 	}
 	if g.update {
 		// TODO: implement update.
 		panic(fmt.Sprintf(
-			"TODO: '%s integration github-app --update'", constants.AppName,
+			"TODO: '%s integration github --update'", constants.AppName,
```
Roming22, line 85: hack/deploy.sh needs to be updated as it is impacted by this change.

otaviof: We don't need to, given we have a alias in place, so the script/users could rely on `github-app` and `tssc integration github` interchangeably.

PR #1250, opened 2025-09-18
Fix RBAC when authProvider is gitlab and host is gitlab.com

dperaza4dustbit:
When working with self managed gitlab instance running under OpenShift the gitlab auth provider worked well. However when using gitlab.com it was not working. It turns out there are config difference depending on wether you use SaaS gitlab or self-managed gitlab. 

## Summary by CodeRabbit

* **New Features**
  * Automatic GitLab host detection to tailor configuration for gitlab.com vs self-managed instances.
  * For gitlab.com: automatically includes users without seat and streamlines setup by omitting the API base URL.
  * For self-managed GitLab: automatically sets the API base URL for proper integration.


Roming22 (approved):
/lgtm

dperaza4dustbit:
re-test

PR #1278, opened 2025-09-25
RHTAP-5637: Deployment Tool Flags

otaviof:
Passing the flags, force, debug and dry-run to the MCP tool, so the installer job uses the same settings. Also, adapting the `tssc_status` tool to recognise a dry-run job.

## Summary by CodeRabbit

* New Features
  * Added support for --dry-run, --debug, and --force flags in the deployment tool.
  * Enforced single-deployment-per-cluster with an option to force-redeploy existing jobs.
  * Introduced pre-deployment topology validation with actionable guidance on failures.
* Documentation
  * Updated help text to explain new flags, dry-run behavior, defaults, and monitoring options.
  * Improved error messages with clearer next steps and a logs command for troubleshooting.

### DIFF
--- a/pkg/mcptools/deploytools.go
+++ b/pkg/mcptools/deploytools.go
@@ -25,8 +25,17 @@ type DeployTools struct {
 
 var _ Interface = &DeployTools{}
 
-// DeployToolName deploy tool name.
-const DeployToolName = constants.AppName + "_deploy"
+const (
+	// DeployToolName deploy tool name.
+	DeployToolName = constants.AppName + "_deploy"
+
+	// DebugArg enables debug mode for the deployment job.
+	DebugArg = "debug"
+	// DryRunArg runs the deployment job on dry-run.
+	DryRunArg = "dry-run"
+	// ForceArg forces the recreation of the deployment job.
+	ForceArg = "force"
+)
 
 // deployHandler handles the deployment of TSSC components.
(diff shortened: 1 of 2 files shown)


dperaza4dustbit (approved):
/lgtm

pkg/mcptools/deploytools.go
```diff
@@ -69,8 +123,35 @@ func (d *DeployTools) Init(mcpServer *server.MCPServer) {
 		Tool: mcp.NewTool(
 			DeployToolName,
 			mcp.WithDescription(`
-Deploys TSSC components to the cluster, uses the cluster configuration to deploy
-the TSSC components sequentially.`,
+Deploys TSSC components to the cluster, using the cluster configuration to deploy
+the components sequentially. Note the "dry-run" flag: the deployment process will
+only be initiated when the "dry-run" flag is set to "false". By default, this flag
```
Roming22, line 128: I find it strange that dry-run is the default. I can't think of another tools that is setup this way.

otaviof: Yes, that's true. Given the nature of platform engineering, the idea is to have the user explicitly confirm that the deployment should happen. Let's validate this idea and change it if we need to.

otaviof: /cc @lcarva

PR #1279, opened 2025-09-25
RHTAP-5695 switch SA for nested pipelines

rhopp:
## Summary by CodeRabbit

- **Tests**
  - Pipeline start invocations now include a service account parameter, running under "konflux-integration-runner".
  - Execution behavior and control flow are unchanged aside from the authenticated identity used during runs.
  - Two test tasks have been updated to use revised upstream task sources and revisions for integration runs.

### DIFF
--- a/integration-tests/pipelines/e2e-main-pipeline.yaml
+++ b/integration-tests/pipelines/e2e-main-pipeline.yaml
@@ -64,9 +64,9 @@ spec:
         resolver: git
         params:
           - name: url
-            value: https://github.com/redhat-appstudio/tssc-cli.git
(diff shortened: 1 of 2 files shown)


rhopp:
Let's hold this one... I just thought this might not be the way to go and that this is "bug" on konflux side. I've asked in https://redhat-internal.slack.com/archives/C04GVLR0155/p1758815883768999. Let's wait for answer...

rhopp:
Closing this one. It was confirmed in slack this will be handled on konflux side.

rhopp:
Konflux chose to use `default` SA, which does not have the RBAC rules we need. Let's switch to `konflux-integration-runner`

PR #1292, opened 2025-09-29
fix(RHTAP-5642): tssc application should support tssc --version to display the application version

raisavarghese:
…splay application version

- Added the flag --version as Global flags
- declared the variable Version in constant package so that flags.go and root.go can access it directly
- version is calculated from git command to list latest tag
- Updated Makefile to inject version via ldflags
- Version injection configured in goreleaser.yaml to support different build methods

## Summary by CodeRabbit

- **New Features**
  - Added a global --version flag that prints the application name and the embedded build version.

- **Chores**
  - Build/release updated to embed a VERSION into the binary at build time via linker flags.
  - VERSION is sourced automatically from Git or CI (with a v0.0.0-SNAPSHOT fallback) and exposed for build-time use.


raisavarghese:
@otaviof ,Regarding the suggestion https://github.com/redhat-appstudio/tssc-cli/pull/1292#pullrequestreview-3304156012 ,If i change os.Exit(0) to return nil.Then tssc --version will always print version number and the help text since the code logic always wanted to show help text if no subcommand is present.Can u pls guide me on this?

pkg/cmd/root.go
```diff
@@ -24,6 +24,18 @@ type RootCmd struct {
 // Cmd exposes the root command, while instantiating the subcommand and their
 // requirements.
 func (r *RootCmd) Cmd() *cobra.Command {
+	// Handle --version globally so it works with subcommands, too.
```
dperaza4dustbit: It makes sense for options like dry-run and help to run on subcommands, but what would version do if you make a call like tssc deploy --version?
I would just make version a subcommand itself so that there is not confusion about the scope of --version when called with subcommand

raisavarghese: Thank you for your feedback.

Initially the flag was not used to work with sub-command,but later was added due to coderabbitai suggestions.Currently the command output is as follows

ravarghe@ravarghe-thinkpadp16vgen1:~/rhtap-product-testing/tssc-cli$ bin/tssc --version
tssc v1.7-freeze-220-g42b3a06d

ravarghe@ravarghe-thinkpadp16vgen1:~/rhtap-product-testing/tssc-cli$ bin/tssc deploy --version
tssc v1.7-freeze-220-g42b3a06d

Please let me know your thoughts if the version as subcommand is a better solution
cc @otaviof @jkopriva @dperaza4dustbit

dperaza4dustbit: if you change version to a subcommand they you would just call tssc version and that is it, no need to figure out what --version with a subcommand means. Just my two cents maybe others have diff experience. 
BTW what does all those other things after v1.7 mean should we just output tssc v1.7? Maybe just distinguish between release and pre-release

dperaza4dustbit: having version as a subcommand is not uncommon btw. Check oc as an example
dperaza@dperaza-mac tssc-cli % oc version
Client Version: 4.18.6
Kustomize Version: v5.4.2

dperaza4dustbit (commented):
Some more points in favor of using sub-command

pkg/cmd/root.go
dperaza4dustbit: if you decide to go with --version maybe handle it like docker cli does

dperaza@dperaza-mac tssc-cli % docker --version
Docker version 28.0.1, build 068a01e
dperaza@dperaza-mac tssc-cli % docker images --version
unknown flag: --version

Usage:  docker images [OPTIONS] [REPOSITORY[:TAG]]

Run 'docker images --help' for more information

jkopriva: I would go with the subcommand(tssc version) instead of flag. As @dperaza4dustbit noted it is more closer to oc, helm...binaries. Running flag --version on deploy does not make sense and we are using the flags for setting parameters for integrations.

Makefile
```diff
@@ -50,6 +50,12 @@ INSTALLER_TARBALL_DATA ?= $(shell find -L $(INSTALLER_DIR) -type f \
 	! -name embed.go \
 )
 
+# Version will be set at build time via git describe
+VERSION ?= $(shell \
+	if [ -n "$(GITHUB_REF_NAME)" ]; then echo "$${GITHUB_REF_NAME}"; \
```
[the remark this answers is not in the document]
xinredhat: I think this is a valid comment

pkg/cmd/root.go
otaviof: > if you decide to go with --version maybe handle it like docker cli does
> 
> dperaza@dperaza-mac tssc-cli % docker --version Docker version 28.0.1, build 068a01e dperaza@dperaza-mac tssc-cli % docker images --version unknown flag: --version
> 
> Usage: docker images [OPTIONS] [REPOSITORY[:TAG]]
> 
> Run 'docker images --help' for more information

I think the idea of handling it as the `docker` CLI does is a good idea. It makes sense to keep it only on the root subcommand, looking at the examples out there, i.e., `tssc --version`.

However, I'd start with the `tssc --version` approach; it's commonly used, and this PR is already doing so. We could evaluate having this as `tssc version` later on, but the `--version` is a good start.

otaviof: And by the way, the version must be able to pin point the exact repository reference, commit-id, the changes are coming from. Let's take `goreleaser` as example to be able to output something like:

```
$ tssc --version
v1.7.0-deadbeef
```

Or, when built from `main`:

```
$ tssc --version
v0.0.0-snapshot-deadbeef
```

It's utterly important to have Git as the source-of-authority in all use-cases.

raisavarghese: @otaviof Thanks for your feedback.As per discussion,the subcommand approach can be taken as an improvement and worked upon later.

Roming22: I agree that it shoud be a subcommand and not an argument. @raisavarghese Please update the PR to reflect that decision.

Roming22:
After reviewing our conversations, I agree that this should have been a subcommand and not a flag. However since the functionality was non-existent and is now implemented, I'm going to take it as a step in the right direction and allow the merge of the PR.

I will be creating a new ticket to switch to a subcommand.

pkg/cmd/root.go
Roming22: After discussing the issue with the team, it was decided that the PR would be merged, and that a `version` subcommand would be implemented at a later date.

Roming22 (changes requested):
There's a couple of AI generated comments that should be processed before the PR can be merged.

raisavarghese:
ravarghe@ravarghe-thinkpadp16vgen1:~/tssc-cli$ bin/tssc --version
**tssc v1.7-freeze-249-g0349d24e**

ravarghe@ravarghe-thinkpadp16vgen1:~/tssc-cli$ bin/tssc 

Trusted Software Supply Chain CLI

Usage:
  tssc [flags]
  tssc [command]

Available Commands:
  completion  Generate the autocompletion script for the specified shell
  config      Manages installer's cluster configuration
  deploy      Rollout TSSC platform components
  help        Help about any command
  installer   Lists or extract the embedded installer resources
  integration Configures an external service provider for TSSC
  mcp-server  Starts the MCP server
  template    Render Helm chart templates
  topology    Shows the installer topology

Flags:
      --debug                   enable debug mode
      --dry-run                 enable dry-run mode
  -h, --help                    help for tssc
      --kube-config string      Path to the 'kubeconfig' file (default "/home/ravarghe/.kube/config")
      --log-level slog.level    log verbosity level (default "warn")
      --timeout time.Duration   helm client timeout duration (default "15m0s")
      **--version                 show the application version**

Use "tssc [command] --help" for more information about a command.

Roming22 (approved):
/lgtm

raisavarghese:
If binary is build by goreleaser:
tssc-cli$ bin/dist/tssc_linux_amd64_v1/tssc --version
tssc 1.7-freeze-SNAPSHOT-102d5a1a

If binary is build by make build command:
tssc-cli$ bin/tssc --version
tssc v1.7-freeze-255-g102d5a1a

Version --> v1.7-freeze
commits advanced from the last tag --> 255
git commit id --> g102d5a1a

In both cases Version number and commit id is same.The formatting can be done in the improvement Jira once we fix the format we needed for tssc version number.

PR #1334, opened 2025-10-07
fix: set DH container resources

Roming22:
A user on a slower system reported that the default values were preventing DH to come up. Those new values fixed the issue.

## Summary by CodeRabbit

* **Chores**
  * Configured CPU request (250m) and CPU limit (4 cores) for the backstage-backend container in Kubernetes deployments.
  * Improves workload predictability and reduces CPU contention for backend services.
  * No direct user-facing changes; operators will see more consistent performance and clearer capacity planning after deploying the updated chart.

### DIFF
--- a/installer/charts/tssc-dh/templates/backstage.yaml
+++ b/installer/charts/tssc-dh/templates/backstage.yaml
@@ -23,6 +23,13 @@ spec:
       spec:
         template:
           spec:
+            containers:
+              - name: backstage-backend
+                resources:
+                  requests:
+                    cpu: 250m
+                  limits:
+                    cpu: '4'
             volumes:
               - $patch: replace
                 name: dynamic-plugins-root


installer/charts/tssc-dh/templates/backstage.yaml
```diff
@@ -23,6 +23,13 @@ spec:
       spec:
         template:
           spec:
+            containers:
+              - name: backstage-backend
+                resources:
+                  requests:
+                    cpu: 250m
```
dperaza4dustbit, line 30: Not a fan of hardcoding resources params, but rather provide defaults and allow user to determine their own size, can this be moved to config file or at least values file? Can you think about the other way, how do we know that this is not too restrictive? Would 4 CPUs be a better limit? What about the memory?

Roming22: User told me the default was a max of 1 CPU. Bumping to 2 seems reasonable until we hear it needs to be bumped up, and the user usecase seemed reasonable and universal to me too (they used a less powerful cluster, and need to use a bit more CPU when the pod is starting up so it doesn't get killed because it's not healthy fast enough).

Sure, we could make that setting visible to the user. But the goal of the CLI is to be able to be used without having to make hard decisions, and limiting the number of things that they have to worry about. So I'm not favorable to that solution at the moment.

If a user really need to heavily customize the settings in a way that's unique to them, I think we would recommend they fork the repo, edit the settings and build their own binary.

dperaza4dustbit: I'm just not convinced what is a good opinionated recommendation, as there is no min hardware or even container requirements in backstage docs. Where was a max of 1 CPU coming from? If we have a level of confidence on our opinion I would agree that less choices is good. But when we are not sure, we should provide a choice to avoid a rebuild if we mis configure our defaults. If we recommend user to fork and provide other defaults, lets at least provide those values in values.yaml rather than hardcoding in template.

dperaza4dustbit: Found some DH recommendations https://docs.redhat.com/en/documentation/red_hat_developer_hub/1.7/html/about_red_hat_developer_hub/rhdh-sizing_about-rhdh

dperaza4dustbit: Also here: https://docs.redhat.com/en/documentation/red_hat_developer_hub/1.7/html/configuring_red_hat_developer_hub/configuring-the-deployment

So the request seems reasonable

dperaza4dustbit: I would go to 4 as the limit

Roming22: The limit is added by the Backstage controller, with a default of 1 CPU.

I will push back on adding settings to values.yaml. The more we expose settings/choices to the user, the less value the installer has as a "let me have a test env without me needing to know anything".

Roming22: CPU limit bumped to 4 as per the DH recommendations.

dperaza4dustbit: Sound Good

dperaza4dustbit (approved):
/lgtm

PR #1370, opened 2025-10-14
RHTAP-5316: Granular set of MCP Configuration Tools

otaviof:
Incorporating the changes from #1276 and continue on the MCP configuration tools.

## Summary by CodeRabbit

- **New Features**
  - Interactive configuration tools: init, global settings, and per-product enable/namespace/properties updates; supports nested key updates and in-place product edits.
  - New utilities for flattening and manipulating configuration values.

- **Bug Fixes / Improvements**
  - Improved YAML read/update behavior preserving structure and anchors.
  - Deploy tooling now uses topology information during setup.

- **Documentation**
  - Added a “Configuration Tools” guide and clarified Phase 2 inspection notes.

- **Chore**
  - Updated dependency declarations and replacements.

- **Tests**
  - Expanded tests for namespace, settings, products, and map-flattening.

### DIFF
--- a/pkg/mcptools/configtools.go
+++ b/pkg/mcptools/configtools.go
@@ -37,6 +37,8 @@ const (
 	NamespaceArg = "namespace"
 	// SettingsArg settings argument.
 	SettingsArg = "setting"
+	// ProductsArg products argument.
+	ProductsArg = "product"
 )
 
 // getHandler similar to "tssc config --get" subcommand it returns a existing TSSC
@@ -98,10 +100,22 @@ func (c *ConfigTools) createHandler(
 	// Setting the namespace from user input, if provided.
 	if ns, ok := ctr.GetArguments()[NamespaceArg].(string); ok {
 		cfg.Installer.Namespace = ns
+		// Set namespace to value of ns
+		if err := cfg.Set("tssc.namespace", ns); err != nil {
+			return nil, err
+		}
 	}
 
-	if settings, ok := ctr.GetArguments()[SettingsArg].(config.Settings); ok {
-		cfg.Installer.Settings = settings
+	if settings, ok := ctr.GetArguments()[SettingsArg].(map[string]interface{}); ok {
+		if err := cfg.Set("tssc.settings", settings); err != nil {
+			return nil, err
+		}
+	}
+
+	if products, ok := ctr.GetArguments()[ProductsArg].(map[string]interface{}); ok {
+		if err := cfg.Set("tssc.products", products); err != nil {
+			return nil, err
+		}
 	}
 
 	// Ensure the configuration is valid.
@@ -173,6 +187,13 @@ and other fundamental services will be deployed.`,
 				mcp.Description(`
 The global settings object for TSSC ('.tssc.settings{}'). When empty the default
 settings will be used.
+				`),
+			),
+			mcp.WithObject(
+				ProductsArg,
+				mcp.Description(`
+The settings object for TSSC ('.tssc.products{}'). When empty the default config
+will be used.
 				`),
 			),
 		),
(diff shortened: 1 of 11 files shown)


pkg/mcptools/configtools.go
```diff
@@ -121,24 +136,293 @@ func (c *ConfigTools) createHandler(
...
+// configSettingsHandler handles the configuration settings update request. It
+// retrieves the current configuration, merges the provided settings (from the
+// 'settings' argument) into the existing installer settings, and updates the
+// cluster configuration.
+func (c *ConfigTools) configSettingsHandler(
+	ctx context.Context,
+	ctr mcp.CallToolRequest,
+) (*mcp.CallToolResult, error) {
+	// Extracting the "settings" argument.
+	settings, ok := ctr.GetArguments()[SettingsArg].(config.Settings)
```
lingyzhuang: With type `config.Settings` here, I will always get `ok=false` and get error returned. Although Gemini tried different inputs for `tssc_config_settings`, like `{"settings":{"crc":true}}`, `{"settings":{"ci":{"debug": false},"crc":true}}`. So I think we should use `map[string]interface{}` here.

pkg/mcptools/configtools.go
```diff
@@ -121,24 +136,293 @@ func (c *ConfigTools) createHandler(
...
+	}
+	return cfg, nil
+}
+
+// getProduct get the product from the configuration, or returns the MCP error.
+func (c *ConfigTools) getProduct(
+	cfg *config.Config,
+	name string,
+) (*config.Product, *mcp.CallToolResult) {
+	spec, err := cfg.GetProduct(name)
```
lingyzhuang, line 228: `spec` here is not type of `*yaml.Node`, so the updated config (for product) will lose the comments. From my test, the updated config would like (Update the namespace of product `develop hub`, but the comments is missing after updating):  

```
       # Exposures (CVE) to assess their risk profile.  
        - name: Trusted Profile Analyzer 
          enabled: true 
          namespace: tssc-tpa 
          properties:  
            manageSubscription: true 
       - name: Developer Hub 
          enabled: true  
          namespace: test 
```

otaviof: Thanks for pointing that out, I just fixed the issue, please try again.

pkg/mcptools/configtools.go
otaviof: I changed the tool signature to use key and value, please consider the latest changes.

pkg/mcptools/configtools.go
```diff
@@ -121,24 +137,289 @@ func (c *ConfigTools) createHandler(
...
+	ctr mcp.CallToolRequest,
+) (*mcp.CallToolResult, error) {
+	key, ok := ctr.GetArguments()[KeyArg].(string)
+	if !ok || key == "" {
+		return mcp.NewToolResultErrorf(`
+You must inform the %q argument with the name of the attribute to update!`,
+			KeyArg,
+		), nil
+	}
+	value, ok := ctr.GetArguments()[ValueArg].(string)
```
lingyzhuang: GetArguments with type of `string`, so update of settings would like `crc: "true"`, maybe set it to `any` or `bool`?

otaviof: Thanks! This argument should be `any`. Just updated.

otaviof: https://github.com/mark3labs/mcp-go/issues/345

For the time being, the library we use for MCP doesn't support `any`, and the `.tssc.settings` only uses booleans. I'm going to leave this tool using a boolean argument for now.

otaviof: https://github.com/mark3labs/mcp-go/pull/618 -- in the future we should have `WithAny` supported, then we can update this logic to support `any` instead of simply `bool`.

lingyzhuang (approved):

Roming22 (approved):
/lgtm

PR #1382, opened 2025-10-20
Integration Overwrite Error

otaviof:
Avoid error out when the interation is already configured, removing the overwrite error.

## Summary by CodeRabbit

* **Bug Fixes**
  * Integration setup is idempotent: already-configured integrations are now skipped instead of causing an error.
  * Re-running inspection or configuration completes gracefully without failing on previously set integrations.
  * Prevents duplicate processing of integrations that were already present, avoiding redundant configuration attempts.


otaviof:
/hold cancel

Double checked the changes, it works as expected.

Roming22 (approved):
/lgtm

Roming22 (approved):
/lgtm

PR #1387, opened 2025-10-22
fix(RHTAP-5734): Display the commit id in the application version

raisavarghese:
- ldflags added for commit id
- updated Makefile to pass build arguments version and commit
- the command output tssc --version displays both version and commit id
- the tssc mcp-server --help will display default value for image

Assisted-By: Cursor

## Summary by CodeRabbit

* **New Features**
  * Version output now shows Version and Commit as two labeled lines.

* **Behavior Changes**
  * CLI image option is no longer required; a default image tag derived from the build commit is used when omitted.

* **Chores**
  * Builds, Docker images and release workflows now accept and embed commit and version values.
  * CI pipelines updated to derive and propagate revision/version metadata into downstream steps.


raisavarghese:
The output is :
ravarghe@ravarghe-thinkpadp16vgen1:~/tssc-cli$ bin/dist/tssc_linux_amd64_v1/tssc --version
tssc Version: 1.7-freeze-SNAPSHOT-6c0fd091
Commit: `6c0fd0918b9d64d445f5da100866ce175343d059`  

ravarghe@ravarghe-thinkpadp16vgen1:~/tssc-cli$ bin/tssc --version
tssc Version: 1.7-freeze-SNAPSHOT-6c0fd091
Commit: `6c0fd0918b9d64d445f5da100866ce175343d059`

Note:
The "go build" command in the Makefile is replaced by "goreleaser build" inorder to keep the version number consistent in all cases and also this will help to use the Version and commit given by goreleaser which is more reliable.

raisavarghese:
we dont have option to have goreleaser in the CI pipelines doesnt have outside network access.The support for goreleaser has to be taken care in another Jira.This PR should be in draft until the goreleaser tool is available.

raisavarghese:
output

raisavarghese:
image

pkg/subcmd/mcpserver.go
```diff
@@ -37,11 +39,7 @@ Starts the MCP server for the TSSC installer, using STDIO communication.
 // PersistentFlags adds flags to the command.
 func (m *MCPServer) PersistentFlags(cmd *cobra.Command) {
 	p := cmd.PersistentFlags()
-	p.StringVar(&m.image, "image", "", "container image for the installer")
-
-	if err := cmd.MarkPersistentFlagRequired("image"); err != nil {
-		panic(err)
-	}
+	p.StringVar(&m.image, "image", fmt.Sprintf("quay.io/redhat-user-workloads/rhtap-shared-team-tenant/tssc-cli:%s", constants.Commit), "container image for the installer\n")
```
otaviof: When the commit-id is "unknonw" it should use the `latest` tag.

Makefile
```diff
@@ -67,17 +61,18 @@ VERSION ?= $(shell \
 # Builds the application executable with installer resources embedded.
 .PHONY: $(BIN)
 $(BIN): installer-tarball
+$(BIN): tool-goreleaser
 	@echo "# Building '$(BIN)'"
 	@[ -d $(BIN_DIR) ] || mkdir -p $(BIN_DIR)
-	go build -ldflags "-X github.com/redhat-appstudio/tssc-cli/pkg/constants.Version=$(VERSION)" -o $(BIN) $(CMD)
+	goreleaser build --clean --snapshot --single-target --output=$(BIN)
```
otaviof: Let's keep `go build` for the default `Makefile` targets, please.

otaviof (changes requested):

Makefile
raisavarghese: @otaviof 
I replaced the command "go build" to "goreleaser build" because of below reasons

- keep the version and commit id populated from goreleaser which is more reliable.
- Using goreleaser command in all cases keeps the version id and commit id same in all builds made by different makefile targets
- Docker file uses default makefile target to create container image.So default target is the one should be updated.
- go build uses git commands to populate version number and commit id which has slightly different format.

Is there any specific reason or benefits for using go build command?

pkg/subcmd/mcpserver.go
raisavarghese: Done the changes

Makefile
otaviof: The primary argument for using `go build` as the default Makefile target is to maintain simplicity and minimize dependencies for the common development workflow. A developer should be able to clone a repository and compile the code with only the standard Go toolchain installed. Using `goreleaser build` for this basic task introduces an unnecessary external dependency and adds overhead, as GoReleaser is designed for the more complex process of creating a release, not just compiling code. This aligns with the principle of separation of concerns: the default build should be for fast, local development or container image builds, while dedicated targets handle more specialized tasks.

The valid concerns about maintaining consistent version and commit information across all builds are best addressed through `Makefile` variables rather than by standardizing on a release tool. By defining variables in the Makefile that pull version information directly from `git`, you create a single source of truth. This information can then be injected into the binary using Go’s build-time linker flags (`ldflags`). This method is powerful, flexible, and allows you to apply consistent versioning to any build target without coupling the core build process to GoReleaser.

Ultimately, the recommended structure is to use the right tool for the right job. The default target should be a simple `go build` for speed and accessibility. A separate `make release` target should leverage the full power of GoReleaser for creating official, distributable artifacts. Other targets, like `make image`, can then use the shared `ldflags` variables to produce consistently versioned images.

raisavarghese: @otaviof Thanks for the detailed explanation.Please confirm the action plan

1.Create a new target( make image) in Makefile for build creation using goreleaser.
2.The docker file should be updated to use make image target for creating container image and the same image will be uploaded to quay.

otaviof: Sure! 🙂 About the steps you suggest, `1.` [is already in place](https://github.com/redhat-appstudio/tssc-cli/blob/ec758995dca5d63aaacb488275e6ee6992021b42/Makefile#L108). And `2.` we need to revert the changes on the `Makefile` to use the regular `go build`:

```
go build -ldflags "-X github.com/redhat-appstudio/tssc-cli/pkg/constants.Version=$(VERSION)" -o $(BIN) $(CMD)
```

Then, test the `make image` and inspect the CI jobs. We need to use [image build args](https://docs.podman.io/en/v5.3.2/markdown/podman-build.1.html#build-arg-arg-value) carry the version (and commit-id), so we don't need to mount `.git` to build the image.

We'll probably need to overwrite `Makefile` variables in the container image build to carry the version information, i.e.:

```sh
make build VERSION="version"
```

Let me know.

.tekton/tssc-cli-pull-request.yaml
```diff
@@ -176,6 +179,65 @@ spec:
...
+            echo "Computed version: $VERSION"
+            printf "%s" "$VERSION" > $(results.version.path)
+      when:
+      - input: $(tasks.init.results.build)
+        operator: in
+        values:
+        - "true"
+      workspaces:
+      - name: basic-auth
+        workspace: git-auth
```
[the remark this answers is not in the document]
raisavarghese: clone-repository artifact hadn't any git information bundled .hence had to clone the repo with full tags.Currently the scope is for public repository.The unused workspace will be removed.

.tekton/tssc-cli-pull-request.yaml
```diff
@@ -176,6 +180,65 @@ spec:
...
+            echo "Cloning repository..."
+            # Use unshallow clone to get full history and tags for git describe
+            git clone --no-single-branch "$GIT_URL" .
+            
+            # Get version using git describe (same logic as Makefile)
+            # Try to get the closest tag, or use commit SHA as fallback
+            VERSION=$(git describe --tags --always 2>/dev/null || echo "v0.0.0-SNAPSHOT")
+            
+            echo "Computed version: $VERSION"
+            printf "%s" "$VERSION" > $(results.version.path)
```
otaviof, line 232: Does it clone the repository twice? How it work in comparison to the current approach to clone and build the project?

pkg/subcmd/mcpserver.go
```diff
@@ -37,11 +39,13 @@ Starts the MCP server for the TSSC installer, using STDIO communication.
 // PersistentFlags adds flags to the command.
 func (m *MCPServer) PersistentFlags(cmd *cobra.Command) {
 	p := cmd.PersistentFlags()
-	p.StringVar(&m.image, "image", "", "container image for the installer")
-
-	if err := cmd.MarkPersistentFlagRequired("image"); err != nil {
-		panic(err)
+	var defaultImage string
+	if constants.Commit == "unknown" || constants.Commit == "" {
+		defaultImage = "quay.io/redhat-user-workloads/rhtap-shared-team-tenant/tssc-cli:latest"
```
otaviof: Let's use the `:latest` image as the default `m.image` value. You can assign the default value on the `NewMCPServer` function, please consider other examples like this in the project.

go.sum
otaviof, line 1: Please revert the changes on this file.

.tekton/tssc-cli-pull-request.yaml
raisavarghese: i had tried to use the source artifact from the result of clone-repository task but that hadnt any .git info in the folder so i added a new task (clone-and-get-version) for cloning repo with all tags.Then we run git commands to get version number and pass it on to the build-container task of pipeline

otaviof:
For the record, let's check if we can use `{{ Revision }}` Tekton Task param and find a way to pass the tag name directly, instead of cloning the repository twice. For now it's okay to keep as is, so we can release 1.8, but let's include a story to tackle this CI improvement. TiA

raisavarghese:
Hi @otaviof ,As discussed 

1. I have changed the image default value as quay.io/redhat-user-workloads/rhtap-shared-team-tenant/tssc-cli:<commit-id> . But note that currently the images are being pushed to quay.io/redhat-user-workloads/rhtap-shared-team-tenant/tssc-cli:on-pr-<commit-id> .we will have to fix the path of image later for a generic approach

2. The tssc command outputs are displayed as below

[RHTAP-5734.pdf](https://github.com/user-attachments/files/23646982/RHTAP-5734.pdf)

.tekton/tssc-cli-pull-request.yaml
```diff
@@ -206,6 +267,10 @@ spec:
       params:
       - name: IMAGE
         value: $(params.output-image)
+      - name: revision
+        value: $(params.revision)
+      - name: version
+        value: $(tasks.get-version.results.version)
```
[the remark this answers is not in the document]
otaviof: Let's tackle this comment if applicable, it sounds like a good idea to avoid leaving unused parameters behind.

.tekton/tssc-cli-push.yaml
```diff
@@ -173,6 +173,65 @@ spec:
...
+            VERSION=$(git describe --tags --always 2>/dev/null || echo "v0.0.0-SNAPSHOT")
+            
+            printf "%s" "$VERSION" > $(results.version.path)
+            echo "Version: $VERSION"
+            echo "Commit ID: $GIT_REVISION"
+      when:
+      - input: $(tasks.init.results.build)
+        operator: in
+        values:
+        - "true"
```
otaviof, line 234: Let's ensure we have a story to tackle the following:
1. Check whether we can rely on `{{ Revision }}` existing param and improve the exiting Task to inform the tag name. Thus, we can avoid cloning the repository twice.
2. Ensure we remove the duplicated task, or find a way to keep a single task definition.

pkg/subcmd/mcpserver.go
```diff
@@ -107,6 +105,13 @@ func NewMCPServer(
 		cfs:    cfs,
 		kube:   kube,
 	}
+
+	// Set default image based on CommitID
+	if constants.CommitID == "" {
+		m.image = "quay.io/redhat-user-workloads/rhtap-shared-team-tenant/tssc-cli:latest"
+	} else {
+		m.image = fmt.Sprintf("quay.io/redhat-user-workloads/rhtap-shared-team-tenant/tssc-cli:%s", constants.CommitID)
+	}
```
otaviof, line 115:
```go
m.image = "quay.io/redhat-user-workloads/rhtap-shared-team-tenant/tssc-cli"

if constants.CommitID == "" {
	m.image = fmt.Sprintf("%s:lattest", m.image)
} else {
	m.image = fmt.Sprintf("%s:%s", m.image, constants.CommitID)
}
```

go.sum
otaviof: Given you're not modifying Go modules, ensure this file is reverted, or not part of your pull request.

.tekton/tssc-cli-push.yaml
raisavarghese: @otaviof Please find the Jira for the same https://issues.redhat.com/browse/RHTAP-6050

PR #1390, opened 2025-10-24
fix: allow custom port for the GitLab service

Roming22:
cf [RHTAP-5995](https://issues.redhat.com//browse/RHTAP-5995)

## Summary by CodeRabbit

* **New Features**
  * Added a --port option to configure custom ports for GitLab integration.
  * Exposed a base URL for non-gitlab.com hosts and standardized the API base URL to use that value.
  * Port handling improved: when port is 443 it’s treated as standard HTTPS and omitted from generated URLs.

### DIFF
--- a/installer/charts/tssc-dh/templates/app-config-content.yaml
+++ b/installer/charts/tssc-dh/templates/app-config-content.yaml
@@ -175,7 +175,8 @@ integrations:
   gitlab:
     - host: ${GITLAB__HOST}
       {{- if ne ($gitlabSecretData.host | b64dec) "gitlab.com" }}
-      apiBaseUrl: https://${GITLAB__HOST}/api/v4
+      baseUrl: ${GITLAB__URL}
+      apiBaseUrl: ${GITLAB__URL}/api/v4
       {{- end }}
       token: ${GITLAB__TOKEN}
 {{- end }}
--- a/installer/charts/tssc-dh/templates/extra-env.yaml
+++ b/installer/charts/tssc-dh/templates/extra-env.yaml
@@ -77,7 +77,11 @@ data:
     GITLAB__GROUP: "{{ $glSecretData.group }}"
     GITLAB__HOST: {{ $glSecretData.host }}
     GITLAB__TOKEN: "{{ $glSecretData.token }}"
+    {{- if eq ($glSecretData.port | b64dec) "443" }}
     GITLAB__URL: {{ print "https://" ($glSecretData.host | b64dec) | b64enc }}
+    {{- else }}
+    GITLAB__URL: {{ print "https://" ($glSecretData.host | b64dec) ":" ($glSecretData.port | b64dec) | b64enc }}
+    {{- end }}
     GITLAB__USERNAME: {{ $glSecretData.username }}
     {{- if and $glSecretData.clientId $glSecretData.clientSecret }}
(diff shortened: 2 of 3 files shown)


pkg/integration/gitlab.go
```diff
@@ -133,6 +138,7 @@ func (g *GitLab) Data(
 
 	return map[string][]byte{
 		"host":         []byte(g.host),
+		"port":         []byte(strconv.Itoa(g.port)),
```
dperaza4dustbit, line 141: Why having to convert, can you just define it as string

pkg/integration/gitlab.go
```diff
@@ -20,6 +21,7 @@ type GitLab struct {
 
 	insecure  bool   // skip tls verification
 	host      string // gitlab host
+	port      int    // gitlab port
```
dperaza4dustbit, line 24: How does the user know it is optional

installer/charts/tssc-dh/templates/app-config-content.yaml
```diff
@@ -174,8 +174,9 @@ integrations:
   {{- $gitlabSecretData := ($gitlabSecretObj.data | default dict) }}
   gitlab:
     - host: ${GITLAB__HOST}
+      baseUrl: ${GITLAB__URL}
```
dperaza4dustbit: I would move this under the if, no need for this var if we are dealing with gitlab.com, test with gitlab.com if you decide to leave here. We have seen the when running with gitlab.com some optional variables actually caused issues.

dperaza4dustbit (changes requested):
Left some comments

pkg/integration/gitlab.go
Roming22: I'd rather an error is raised if the input is bad.

installer/charts/tssc-dh/templates/app-config-content.yaml
Roming22: Fixed

pkg/integration/gitlab.go
Roming22: We don't flag optional params at the moment. If we want to start doing so, we should open a story for it.

pkg/integration/gitlab.go
dperaza4dustbit: sounds good

pkg/integration/gitlab.go
dperaza4dustbit: I'm OK if we add a story for this

installer/charts/tssc-dh/templates/extra-env.yaml
```diff
@@ -77,7 +77,11 @@ data:
     GITLAB__GROUP: "{{ $glSecretData.group }}"
     GITLAB__HOST: {{ $glSecretData.host }}
     GITLAB__TOKEN: "{{ $glSecretData.token }}"
+    {{- if eq ($glSecretData.port | b64dec) "443" }}
```
dperaza4dustbit, line 80: since the secret will be an integer will this be always false? Not sure, would have to test

Roming22: I tested, and the URL was set properly.

dperaza4dustbit (approved):
/lgtm

PR #1397, opened 2025-10-31
RHTAP-5997: MCP Notes Tool

otaviof:
Shows the connection information for products.

## Summary by CodeRabbit

* **New Features**
  * Added a Notes tool to fetch product-specific deployment NOTES and connection instructions.
  * Added deployment verification with automatic retries and a direct way to retrieve release notes.

* **Improvements**
  * Clearer, phase-aware installer status messages and improved status behavior.
  * More actionable cluster-configuration error text with guided next steps.
  * Exposed topology collection for easier inspection and troubleshooting.

### DIFF
--- a/pkg/deployer/helm.go
+++ b/pkg/deployer/helm.go
@@ -147,6 +147,8 @@ func (h *Helm) Verify() error {
 	return nil
 }
 
+// VerifyWithRetry attempts to verify the Helm deployment multiple times with a
+// delay between retries.
 func (h *Helm) VerifyWithRetry() error {
 	var err error
 	var retries = 3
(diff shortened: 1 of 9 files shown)


pkg/deployer/helm.go
```diff
@@ -178,6 +180,19 @@ func (h *Helm) VisitReleaseResources(
 	})
 }
 
+// GetNotes retrives the latest release (version 0) of the Helm chart, printing
```
Roming22: *retrieves
`latest release` conflicts with `version 0` which is the initial version.

pkg/mcptools/notestool.go
```diff
@@ -0,0 +1,150 @@
...
+// and return the notes.
+func (n *NotesTool) notesHandler(
+	ctx context.Context,
+	ctr mcp.CallToolRequest,
+) (*mcp.CallToolResult, error) {
+	// Ensure the user has provided the product name.
+	name := ctr.GetString(NameArg, "")
+	if name == "" {
+		return mcp.NewToolResultError(`
+		You must inform the Red Hat product name`,
```
Roming22, line 50: So we use the product name instead of the chart name?

pkg/deployer/helm.go
otaviof: The typo is fixed. And, for Helm the latest release means `0`, in this file we have the application's Helm client.

pkg/mcptools/notestool.go
otaviof: Yes, at least initially. Through the MCP server you don't have access to Helm charts, when we do we could get notes for a specific chart name, instead of product name. Perhaps having both tools available.

otaviof: By the way, when we have the `tssc_topology` tool we can expand the scope of the notes' tool and support any Helm chart, as long as we have the topology tool to let the LLM context "know" about the existing names.

Roming22 (approved):
/lgtm

PR #1454, opened 2025-11-12
feat: MCP Server Debug

otaviof:
Adding documentation and scripting to debug the `tssc mcp-server` subcommand using Delve.

## Summary by CodeRabbit

* **Documentation**
  * Added a debugging guide for the MCP server with step‑by‑step instructions, VSCode launch examples, and reference links.

* **Chores**
  * Added a debug build target that produces binaries suitable for debugging.
  * Added a helper script to simplify launching the Delve debugger against the MCP server for local development.

### DIFF
--- a/hack/dlv-tssc-mcp-server.sh
+++ b/hack/dlv-tssc-mcp-server.sh
@@ -1,0 +1,95 @@
+#!/usr/bin/env bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# Script's directory full path.
+declare SCRIPT_DIR
+SCRIPT_DIR="$(dirname "${0}")"
+# Installer executable, by default relative to this script.
+declare TSSC_BIN="${SCRIPT_DIR}/../bin/tssc"
+
+# Installer container image, a required flag for the mcp-server subcommand.
+declare TSSC_IMAGE="ghcr.io/redhat-appstudio/tssc:latest"
+# Port to expose DLV API.
+declare DLV_PORT="8282"
+
+# Shows the script usage and flags.
+usage() {
+    echo "
+This script wraps 'dlv exec' on 'tssc mcp-server', preserves the STDIO
+communication while exposing Delve's API on a local port. The goal is using this
+script as the command for the agentic LLM client while running Delve debugger
+against the CLI.
+
+Usage:
+    ${0##*/} [options]
+
+Optional arguments:
+    -i, --image IMAGE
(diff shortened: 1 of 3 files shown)


hack/dlv-tssc-mcp-server.sh
```diff
@@ -0,0 +1,40 @@
...
+set -o nounset
+set -o pipefail
+
+# Current script directory.
+declare -r SCRIPT_DIR="$(dirname ${0})"
+
+# Installer executable, by default relative to this script.
+declare -r TSSC_BIN="${TSC_BIN:-${SCRIPT_DIR}/../bin/tssc}"
+# Installer container image, a required flag for the mcp-server subcommand.
+declare -r TSSC_IMAGE="${TSSC_IMAGE:-ghcr.io/redhat-appstudio/tssc:latest}"
```
Roming22: Why not give the option to use a local dev image? Look at the usage of `deploy.sh` (`--container`).
I'm also surprised we can use any image when the docs specify the image should be built with debug mode enabled.

hack/dlv-tssc-mcp-server.sh
Roming22, line 1: All the scripts in that folder have a common template. It would be great if this new script also followed that structure.

hack/dlv-tssc-mcp-server.sh
otaviof: To clarify, you only need `make debug` to build the installer executable with debug symbols. The container is only employed to create the deployment job in the cluster; it is not subject to debugging—at least not from the perspective of the `tssc mcp-server`.

Let me amend the script to use the same style as the `deploy.sh --container` on this script too.

By the way, in the near future, this flag `--image` on `tssc mcp-server` will become optional. Users will be encouraged to use the default.

hack/dlv-tssc-mcp-server.sh
```diff
@@ -0,0 +1,96 @@
...
+" >&2
+}
+
+# Parse command-line flags.
+parse_args() {
+    while [[ ${#} -gt 0 ]]; do
+        case ${1} in
+        -d | --debug)
+            set -x
+            shift
```
Roming22: shift not needed here as there is no value to the parameter.

Roming22 (commented):
Minor change needed, but otherwise looks fine.

Roming22 (approved):
/lgtm

PR #1526, opened 2025-11-21
fix: tas integration

Roming22:
cf [RHTAP-6049](https://issues.redhat.com//browse/RHTAP-6049)

## Summary by CodeRabbit

* **New Features**
  * Added Trusted Artifact Signer integration support

* **Changes**
  * Removed support for Developer Hub, OpenShift GitOps, and Trusted Profile Analyzer integrations
  * Updated trusted infrastructure endpoint discovery to use secret-based configuration

### DIFF
--- a/hack/deploy.sh
+++ b/hack/deploy.sh
@@ -25,8 +25,8 @@ Optional arguments:
     -e, --env-file
         Environment variables definitions (default: $SCRIPT_DIR/private.env)
     -i, --integration INTEGRATION
-        Use an external service [acs, bitbucket, ci, dh, github, gitlab, gitops,
-        jenkins, quay, tas, tpa].
+        Use an external service [acs, bitbucket, ci, github, gitlab,
+        jenkins, quay, tas].
     -d, --debug
         Activate tracing/debug mode.
     -h, --help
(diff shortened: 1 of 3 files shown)


dperaza4dustbit (approved):
/lgtm

hack/deploy.sh
```diff
@@ -259,6 +235,11 @@ integrations() {
             --dockerconfigjson='"$QUAY__DOCKERCONFIGJSON"' \
             --token='"$QUAY__API_TOKEN"' --url='"$QUAY__URL"'
     fi
+    if [[ -n "${TAS:-}" ]]; then
+        tssc_cli integration trusted-artifact-signer --force \
+            --rekor-url='"$TAS__REKOR_URL"' \
+            --tuf-url='"$TAS__TUF_URL"'
```
otaviof, line 241: Any specific reason for the single and double quotes together? I think using a simple `"${TAS__REKOR_URL}"` should work as expected.

Roming22: Yes, it has to do with how the binary is then called, which can be either directly or through the image.
I agree it's a little bit to look at though, so if you have a better solution we can change it.

PR #1565, opened 2025-12-02
Remove hardcoded product name

lingyzhuang:
Add function `ProductName` in `integration.go`, get product name from metadata instead of hardcoded.
Jira: [RHTAP-6037](https://issues.redhat.com/browse/RHTAP-6037)

## Summary by CodeRabbit

* **New Features**
  * Integration command now bootstraps configuration at runtime and automatically disables configured integrations based on discovered chart-to-product mappings.

* **Refactor**
  * Individual integration commands no longer perform incidental product enable/disable flows; creation runs are more direct and predictable.
  * Resolver enhanced to map integrations to products for consistent post-setup handling.

* **Chores**
  * Command wiring updated to provide embedded chart filesystem access for integration discovery.


pkg/subcmd/integration.go
```diff
@@ -1,21 +1,106 @@
...
+		return err
+	}
+
+	return config.NewConfigMapManager(kube).Update(ctx, cfg)
+}
+
+func NewIntegration(
+	logger *slog.Logger,
+	kube *k8s.Kube,
+	cfs *chartfs.ChartFS,
```
otaviof, line 18: Make sure the `cfs` is used.

pkg/subcmd/integration.go
```diff
@@ -1,21 +1,106 @@
...
+
+	for _, hc := range charts {
+		d := resolver.NewDependency(&hc)
+		if slices.Contains(d.IntegrationsProvided(), name) {
+			return d.ProductName(), nil
+		}
+	}
+
+	return "", fmt.Errorf("failed to get product name from integration: %s", name)
+}
```
otaviof: This method belongs to the [`resolver.Collection`](https://github.com/redhat-appstudio/tssc-cli/blob/8fa8c77642f320180e2bc18c999e5030a5e21981/pkg/resolver/collection.go), have a look on the other methods in there, please.

pkg/subcmd/integration.go
```diff
@@ -1,21 +1,106 @@
...
+) error {
+	pName, err := productName(name)
+	if err != nil {
+		// Ignore "not found" errors, some integrations do not require updates
+		// in config, e.g. quay
+		if strings.Contains(err.Error(), "failed to get product name from integration") {
+			return nil
+		}
+		return err
+	}
```
otaviof: Here you need to consider the errors defined in the `resolver` package, and instead of using `strings.Contains`, you need to leverage `errors.Is()`. Instead of trying to match the error description, `errors.Is()` allows you to match an error type.

otaviof (changes requested):

pkg/resolver/collection.go
```diff
@@ -73,6 +73,30 @@ func (c *Collection) GetProductDependency(product string) (*Dependency, error) {
...
+// GetProductName searches and returns the product name by integration name.
+// It goes though all charts and search for annotation "integrations-provided"
+// If it matches integration name, then returns product name, which is from
+// annotation "product-name"
+func (c *Collection) GetProductName(integrationName string) (string, error) {
+	var productName string
+	_ = c.Walk(func(_ string, d Dependency) error {
+		if productName != "" {
+			return fmt.Errorf("found") // stop walking
+		}
```
otaviof, line 85: Use `return nil` to short circuit, you should only return error when a error occurs.

pkg/resolver/collection.go
```diff
@@ -73,6 +73,30 @@ func (c *Collection) GetProductDependency(product string) (*Dependency, error) {
 	return productDependency, nil
 }
 
+// GetProductName searches and returns the product name by integration name.
+// It goes though all charts and search for annotation "integrations-provided"
+// If it matches integration name, then returns product name, which is from
+// annotation "product-name"
+func (c *Collection) GetProductName(integrationName string) (string, error) {
```
otaviof: The function name should reflect what it does, my suggestion would be: `GetProductNameForIntegration`. Which reflects the idea, we're trying to identify a product name for the (informed) integration.

pkg/resolver/collection.go
```diff
@@ -73,6 +73,30 @@ func (c *Collection) GetProductDependency(product string) (*Dependency, error) {
...
+		}
+		return nil
+	})
+
+	if productName == "" {
+		return "", ErrUnknownIntegration
+	}
+
+	return productName, nil
+}
```
otaviof: This function should return empty (`func _(integrationName string) string {}`) when a integration is not covered by a product. For instance, `github`, `gitlab`, etc. These integrations aren't associated with integrations.

For the record, the `resolver.Collection` is validated during instantiation, product and integration names are validated:

https://github.com/redhat-appstudio/tssc-cli/blob/2dc1d971afeccdcb32e77b1520297b33748db9be/pkg/resolver/collection.go#L83-L106

pkg/subcmd/integration.go
```diff
@@ -1,21 +1,89 @@
...
+			if err != nil {
+				return err
+			}
+
+			configuredIntegrations, err := manager.ConfiguredIntegrations(cmd.Context(), cfg)
+			if err != nil {
+				return err
+			}
+
+			for _, name := range configuredIntegrations {
```
otaviof: nit: `integrationName` instead of `name` 🙏

pkg/subcmd/integration.go
```diff
@@ -1,21 +1,89 @@
...
+				productName, err := collection.GetProductName(name)
+				if errors.Is(err, resolver.ErrUnknownIntegration) {
+					continue
+				}
+
+				if err := disableProduct(
+					cmd.Context(),
+					productName,
+					cfg,
+					kube,
```
otaviof: nit: perhaps you can keep that in a single line...

Roming22 (approved):

pkg/resolver/collection.go
```diff
@@ -73,6 +73,31 @@ func (c *Collection) GetProductDependency(product string) (*Dependency, error) {
 	return productDependency, nil
 }
 
+// GetProductName searches and returns the product name by integration name.
```
otaviof: Function name needs to be renamed here too.

pkg/subcmd/integration.go
```diff
@@ -1,21 +1,83 @@
...
+	if err != nil {
+		return err
+	}
+
+	spec.Enabled = false
+	if err := cfg.SetProduct(productName, *spec); err != nil {
+		return err
+	}
+
+	return config.NewConfigMapManager(kube).Update(ctx, cfg)
```
otaviof: nit: I'd move these lines into the `PersistentPostRunE` block, given it's only a few lines.

pkg/resolver/collection.go
```diff
@@ -73,6 +73,31 @@ func (c *Collection) GetProductDependency(product string) (*Dependency, error) {
...
+			productName = d.ProductName()
+		}
+		return nil
+	})
+
+	if productName == "" {
+		return ""
+	}
+
+	return productName
```
otaviof: Here you can simply `return productName`, whether the name is empty that's the function return as expected.

pkg/subcmd/integration.go
```diff
@@ -3,19 +3,73 @@ package subcmd
...
+				if err := cfg.SetProduct(productName, *spec); err != nil {
+					return err
+				}
+
+				updated = true
+			}
+
+			if updated {
+				return config.NewConfigMapManager(kube).Update(cmd.Context(), cfg)
+			}
```
otaviof, line 69: Here we should inspect whether `spec.Enabled` is currently enabled, and upon getting disabled we should then trigger the `config.NewConfigMapManager().Update()`.

otaviof (approved):
/lgtm

PR #1570, opened 2025-12-03
fix: use source repo name from PR for rhads-config download

rhopp:
The get-rhads-config task was incorrectly using '.git.repo' field which doesn't exist in the job-spec. This caused issues when forks still use the old 'rhtap-cli' repository name instead of the new 'tssc-cli' name.

Changed to extract the repository name from 'source_repo_url' which correctly reflects the actual fork's repository name, whether it's 'rhtap-cli' or 'tssc-cli'.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/integration-tests/tasks/get-rhads-config.yaml
+++ b/integration-tests/tasks/get-rhads-config.yaml
@@ -28,7 +28,9 @@ spec:
         set -euo pipefail
 
         echo "Downloading rhads-config file:"
-        GIT_REPO="$(jq -r '.git.repo // empty' <<< $JOB_SPEC)"
+        # Extract the repository name from the source_repo_url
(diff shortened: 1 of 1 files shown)


Roming22 (approved):
/lgtm

rhopp:
At least one of the plrs succeeded, but the overall step failed, because the sucessfull pipelinerun was already pruned from the cluster.
This is the PLR that succeeded: https://konflux-ui.apps.stone-prd-rh01.pg1f.p1.openshiftapps.com/ns/rhtap-shared-team-tenant/applications/tssc-cli-1-8/pipelineruns/e2e-4.19-f2g9m

PR #1598, opened 2025-12-19
fix(plugins): conditionally render Gitlab pipelines

CryptoRodeo:
The Gitlab Pipelines card has an issue where it will always try to render in the CI tab.

This change adds a condition to how it's rendered to prevent that. Now in order to render the Gitlab Pipelines card you'll need to use the `gitlab.com/project-id` annotation.

[The templates](https://github.com/redhat-appstudio/tssc-sample-templates/blob/7a948d50565652b0a260b259027095dfa9e6e1f6/skeleton/source-repo/catalog-info.yaml#L20) use `gitlab.com/project-slug` so the Gitlab pipelines should be hidden from the CI tab now.

JIRA: [RHTAP-4055](https://issues.redhat.com//browse/RHTAP-4055)

## Summary by CodeRabbit

* **New Features**
  * Added a GitLab CI/CD pipelines table to entity pages.
  * The table displays full-width and appears only when GitLab integration is available and the entity is configured with the GitLab project identifier.


Roming22 (approved):
/lgtm

PR #1601, opened 2026-01-05
Moving reference for templates to consolidated repo

dperaza4dustbit:
Both tssc-sample-templates and tssc-sample-pipelines have been consolidated under tssc-dev-multi-ci. Changing reference in installer config

## Summary by CodeRabbit

* **Chores**
  * Updated the Developer Hub template catalog location in the installer configuration and related update script to point to a new repository path.

### DIFF
--- a/hack/code-freeze.sh
+++ b/hack/code-freeze.sh
@@ -102,7 +102,7 @@ update_charts() {
 update_template() {
     CONFIG="installer/config.yaml"
     get_version "developerHub"
-    CATALOG_URL="https://github.com/redhat-appstudio/tssc-sample-templates/blob/release-v${VERSION_XY}.x/all.yaml" \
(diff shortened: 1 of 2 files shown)


Roming22 (changes requested):
@dperaza4dustbit We usually don't link to `main` to prevent changes to the templates to impact the CLI CI.

My preference would be to create a branch or a tag and use that instead.

You also need to update the repo url in `hack/code-freeze.sh`.

dperaza4dustbit:
> @dperaza4dustbit We usually don't link to `main` to prevent changes to the templates to impact the CLI CI.
> 
> My preference would be to create a branch or a tag and use that instead.
> 
> You also need to update the repo url in `hack/code-freeze.sh`.

Done, PTAL

Roming22 (approved):
/lgtm

PR #1618, opened 2026-01-26
Konflux integration improvments

gbenhaim:
- Enable image controller and the creation of a quay token for it
- Create a secret with the public key chains is using, and give authenticated users permission to see chains public key. required by pipelines which runs Conforma.
- Inject github app url to the konflux config so it would be presented in konflux UI.

### DIFF
--- a/installer/charts/tssc-konflux/templates/job-public-key.yaml
+++ b/installer/charts/tssc-konflux/templates/job-public-key.yaml
@@ -1,0 +1,64 @@
+{{- $pipelinesNamespace := .Values.pipelines.namespace }}
+{{- $signingSecretName := "signing-secrets" -}}
+{{- $publicKeySecretName := "public-key" -}}
+{{- $signingSecretObj := (
+      lookup "v1" "Secret" $pipelinesNamespace $signingSecretName
+    ) | default dict
+-}}
+{{- $signingSecretData := (get $signingSecretObj "data") | default dict -}}
+{{- $signingCosignPubB64 := (get $signingSecretData "cosign.pub") | default "" -}}
+{{- $signingCosignPub := ternary ($signingCosignPubB64 | b64dec) "" (ne $signingCosignPubB64 "") -}}
+{{- $publicKeySecretObj := (
+      lookup "v1" "Secret" $pipelinesNamespace $publicKeySecretName
+    ) | default dict
+-}}
+{{- $publicKeySecretData := (get $publicKeySecretObj "data") | default dict -}}
+{{- $publicKeyCosignPubB64 := (get $publicKeySecretData "cosign.pub") | default "" -}}
+{{- $publicKeyCosignPub := ternary ($publicKeyCosignPubB64 | b64dec) "" (ne $publicKeyCosignPubB64 "") -}}
+{{- if and (ne $signingCosignPub "") (ne $signingCosignPub $publicKeyCosignPub) }}
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  annotations:
+    helm.sh/hook: post-install,post-upgrade
+  labels:
+    {{- include "common.labels" . | nindent 4 }}
+  name: tssc-konflux-public-key
+spec:
+  template:
+    spec:
+      serviceAccountName: {{ .Release.Name }}
+      restartPolicy: Never
+      containers:
+        - name: copy-public-key
+          image: quay.io/codeready-toolchain/oc-client-base:latest
+          command:
+          - /bin/bash
+          - -c
+          - |
+            set -o errexit
+            set -o nounset
+            set -o pipefail
+
+            PUBLIC_KEY=${0:?Public key is required}
+            PIPELINES_NAMESPACE=${1:?Pipelines namespace is required}
+            PUBLIC_KEY_SECRET_NAME="public-key"
+
+            echo "Deleting existing public-key secret if it exists"
+            oc delete secret -n "$PIPELINES_NAMESPACE" "$PUBLIC_KEY_SECRET_NAME" || :
+
+            echo "Creating public-key secret in $PIPELINES_NAMESPACE namespace"
+            oc create secret generic \
+              "$PUBLIC_KEY_SECRET_NAME" \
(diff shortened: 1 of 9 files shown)


installer/charts/tssc-pipelines/templates/job-tekton-chains.yaml
```diff
@@ -30,10 +47,38 @@ spec:
...
+            set -o errexit
+            set -o nounset
+            set -o pipefail
+
+            SECRET_NAME=${0:?Secret name is required}
+            NAMESPACE=${1:?Namespace is required}
+            SECRET_URI="k8s://${NAMESPACE}/${SECRET_NAME}"
+            PUBLIC_KEY_SECRET_NAME="public-key"
+
+            echo "Cleaning up existing secrets"
```
Roming22: Charts should be idempotent. If keys are already there, they should not be deleted and regenerated.
I would advise to remove lines 62-65, and gate lines 67-69 with `{{- if eq $cosignKey "" }}`.

Note that this comment can be ignored in favor of https://github.com/redhat-appstudio/tssc-cli/pull/1618/changes#r2728080697.

installer/charts/tssc-pipelines/templates/job-tekton-chains.yaml
```diff
@@ -30,10 +47,38 @@ spec:
...
+
+            /workspace/oc delete secret -n "$NAMESPACE" "$SECRET_NAME" || :
+            /workspace/oc delete secret -n "$NAMESPACE" "$PUBLIC_KEY_SECRET_NAME" || :
+
+            echo "Generating new key pair $SECRET_URI"
+
+            cosign generate-key-pair "$SECRET_URI"
+
+            PUBLIC_KEY=$(cosign public-key --key "$SECRET_URI")
```
Roming22: This `public-key` secret seems to be relevant only to Konflux. So I would advise the following:

- Revert changes to that file
- Add a dependency to `tssc-pipelines` in `tssc-konflux/Chart.yaml` under `tssc.redhat-appstudio.github.com/depends-on:`.
- Add a job to create the `public-key` secret in the tssc-konflux chart.

pkg/subcmd/integration_konflux_quaytoken.go
Roming22, line 1: Integration should be generic. Can you please extend the existing `pkg/subcmd/integration_quay.go` command so it can handle your usecase?

That might require making some parameters optional and adding some `{{- require }}` in the chart.

installer/charts/tssc-pipelines/templates/job-tekton-chains.yaml
gbenhaim: I thought about this, but it means that the konflux chart would create a resource in the openshift-pipelines namespace which didn't feel right to me. I wouldn't say that the existence of a secret with the public key is konflux specific. Without it, how anyone can verify the signature of attestations?

pkg/subcmd/integration_konflux_quaytoken.go
gbenhaim: I saw integrations that are specific to acs, rhtas, jeninks... how is it different? I think it would be better to add a new integration instead of making an existing one more complex.

installer/charts/tssc-pipelines/templates/job-tekton-chains.yaml
gbenhaim: If the keys are already there the Job won't be created. See the condition above the secret resource definition.

Roming22: If `signing-secrets` exists, but `public-key` does not, `signing-secrets` would be rotated, which is not the desired behavior.

installer/charts/tssc-pipelines/templates/job-tekton-chains.yaml
Roming22: I agree that it feels wrong. However the name and location of that secret is controlled by Konflux (if I recall correctly, we were the one that chose the name and location of that secret). As long as that is true, it needs to be managed by the Konflux chart.

I would be fine with any of the 3 following options:

1. Move the creation of `public-key` to the `tssc-konflux` chart.
2. Move the creation of `public-key` to the `tssc-konflux` chart, rename the secret and move it to a Konflux owned namespace.
3. Have Tekton Chains create the manage the `public-key` secret, thus making it the official way to access the information.

pkg/subcmd/integration_konflux_quaytoken.go
Roming22: There should be one integration per service. What you are doing is creating a new integration to the quay service, which we're trying to discourage.

gbenhaim: done

installer/charts/tssc-pipelines/templates/job-tekton-chains.yaml
gbenhaim: I moved the creation of public-key to the tssc-konflux chart.

installer/charts/tssc-pipelines/templates/job-tekton-chains.yaml
gbenhaim: I dropped the changes to this file

installer/config.yaml
Roming22, line 1: Why is that change needed?

installer/charts/tssc-konflux/templates/job-public-key.yaml
Roming22, line 1: Cool. Now you have a dependency on the `tssc-pipelines` chart, since it's responsible for the creation of `signing-secrets`. You need to declare it in `installer/charts/tssc-konflux/Chart.yaml` so charts get installed in the right order.

Roming22 (changes requested):
Almost there ;)

installer/config.yaml
Roming22: Ah, it's because of a known bug [RHTAP-6105](https://issues.redhat.com/browse/RHTAP-6105).

Roming22 (approved):
/lgtm

PR #1619, opened 2026-01-26
fix: GitOps could not be deployed without RHDH

Roming22:
rh-pre-commit.version: 2.3.2


xinredhat (commented):
/lgtm

PR #1625, opened 2026-01-30
fix: Make the config options obvious in the error message

ralphbean:
A new user running `tssc config` for the first time is greeted with an error message that guides them to try `tssc config create` next, but the right thing to do is `tssc config --create`.

Update the error message to help the user do the right thing the first time.


Roming22 (approved):
/lgtm

PR #1627, opened 2026-02-02
Adding explicit timeouts instead of using konflux defaults

dperaza4dustbit:
We are having issue with e2e sub-pipeline runs timing out after exactly 1 hour. Adding timeouts to override Konflux defaults that seem to have changed. 

## Summary by CodeRabbit

* **Chores**
  * Extended nested pipeline execution timeout to 2h30m to reduce unexpected terminations.
  * Added a pipeline start timeout parameter to allow specifying runtime timeouts when launching pipelines.
  * Added a post-start verification step that reads and logs the created PipelineRun timeout for easier troubleshooting.

### DIFF
--- a/integration-tests/pipelines/e2e-main-pipeline.yaml
+++ b/integration-tests/pipelines/e2e-main-pipeline.yaml
@@ -57,6 +57,7 @@ spec:
         - name: job-spec
           value: $(tasks.test-metadata.results.job-spec)
     - name: start-nested-pipelines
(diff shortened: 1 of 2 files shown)


Roming22 (approved):
/lgtm

dperaza4dustbit:
Going to force merge since this will never pass due to the timeout default of 1 hour. I was able to point to my fork and see the timeout issue resolved

PR #1631, opened 2026-02-04
fix(dh): remove GITHUB__TOKEN from the DH config

Roming22:
This change removes the GITHUB__TOKEN from the DH config to prevent the token from being exposed in the DH config, and removes the GITHUB__USERNAME variable as it is not used.

This change is part of the fix for the GH integration issue.

cf [RHTAP-6135](https://issues.redhat.com//browse/RHTAP-6135)

## Summary by CodeRabbit

* **Chores**
  * Removed the GitHub integration token from application configuration templates so credentials are no longer emitted by default.
  * Removed GitHub authentication environment variables (token and username) from deployment configuration emissions to reduce exposed secrets during deployment.


dperaza4dustbit (approved):
/lgtm

PR #1638, opened 2026-02-11
Add sigstore configuration to Konflux CR

gbenhaim:
Populate the Konflux CR's clusterConfig with sigstore URLs (Fulcio, Rekor, TUF) by looking up Ingress and Service resources from the Trusted Artifact Signer namespace at deploy time.

External URLs are derived from Ingress hosts, internal URLs from Service metadata. All lookups use Helm's `required` function to fail early with a clear error if any resource is missing.

Assisted-By: Cursor


Roming22 (approved):
/lgtm

PR #1641, opened 2026-02-12
fix: use correct k8s service DNS names in TAS integration secret

gbenhaim:
The internal service URLs were incorrectly embedding the namespace into the service name (e.g. fulcio-server-<ns>.<ns>.svc), instead of using the proper fully qualified in-cluster DNS format (<service>.<namespace>.svc.cluster.local).

Assisted-By: Cursor


Roming22 (approved):
/lgtm

PR #1778, opened 2026-04-07
Fix: keycloak route

smatula:
New clusters offered by the Red Hat Catalog now have a default route that is conflicting with tssc.  Adding a prefix to fix the collision.

## Summary by CodeRabbit

* **Configuration Updates**
  * Updated Keycloak SSO endpoint hostname from `sso` to `tssc-sso` subdomain. Dependent authentication URLs and issuer configurations reflect this change.


Roming22 (approved):
/lgtm

PR #1779, opened 2026-04-13
Fix TPA login via keycloak

dperaza4dustbit:
Keycloak 26 introduces more requirements of client protocol mappers. This fixes the frontend client being used by TAP

If more client need this we might need to move this protocol mapper to the profile scope [protocol mapper](https://github.com/redhat-appstudio/tssc-cli/blob/main/installer/charts/tssc-iam/templates/realm.yaml#L662)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed identity claim handling in the authentication system by explicitly configuring how user identity attributes map to standard identity claims. The mapping is now properly emitted across all token types including ID tokens, access tokens, userinfo responses, and token introspection operations, ensuring consistent identity information handling throughout authentication and authorization flows.

### DIFF
--- a/installer/charts/tssc-iam/templates/realm.yaml
+++ b/installer/charts/tssc-iam/templates/realm.yaml
@@ -1419,6 +1419,20 @@ spec:
           - offline_access
           - microprofile-jwt
         protocol: openid-connect
+        protocolMappers:
+          - name: "subject-mapper"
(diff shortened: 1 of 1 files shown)


Roming22 (commented):
/lgtm

